PR #109, opened 2015-01-23
Make tests run in lua5.1 by removing ffi dep

avidal:
Windows detection was being done with luajit ffi module, which isn't available
in lua5.1. Now all tests will pass with lua5.1.

Maybe lua5.1 should be added to the travis config?


tests/platform.lua
```diff
@@ -0,0 +1,34 @@
...
+if system == "Darwin" then
+  platform.unix = true
+  platform.macosx = true
+elseif system == "Linux" then
+  platform.unix = true
+  platform.linux = true
+elseif system and system:match("^CYGWIN") then
+  platform.unix = true
+  platform.cygwin = true
+elseif system and system:match("^Windows") then
```
justinmk: Windows doesn't have uname.

tests/platform.lua
```diff
@@ -0,0 +1,34 @@
+-- Determine some platform information used for tests
+-- Adapted from: https://github.com/keplerproject/luarocks/blob/master/src/luarocks/cfg.lua
+local platform = {}
```
justinmk: Using a dict with keys for each platform is a weird approach...

tests/platform.lua
avidal: Yeah, I thought the same thing but don't have a Windows box to verify. These checks are taken from the luarocks cfg module. Cygwin and Mingw at least expose uname and these seem to check out for those environments.

I suppose it can be adjusted to use `ver` or `winver` or some other check for Windows (systeminfo?) but I don't have a Windows box to test it against.

tests/platform.lua
avidal: Yeah, true. Again, I "borrowed" the implementation from luarocks.cfg; but they likely use the combination of platforms for various things, whereas the luv tests only really need to know if the os is Windows or not.

creationix:
Thanks, but I found a less intrusive way to detect the platform.

PR #114, opened 2015-02-16
fix(timer): update the time on luv_now

rphillips:


creationix:
While this does take away the hazard of using `uv_now` incorrectly, it makes it impossible to use `uv_update_timer` correctly in many cases (causing it to be called too often and slowing performance).

Where is the code that's using `uv_now` and not getting updated times?

rphillips:
I am thinking this might be the issue with the random failures we are seeing with the socket timeout test. The timer code in luvit uses uv.now() to calculate certain values... It isn't a smoking gun, just an observation.

PR #119, opened 2015-03-13
fix lua 5.2 load error because without `lua_objlen`

AndrewTsao:


creationix:
I test against luajit which is mostly lua 5.1 style. Thanks for the patch.

PR #121, opened 2015-03-17
propogate raw status

rphillips:


creationix:
What's this for?

creationix:
Also the existing `uv_err_name()` return value should match 1-1 with the status code.  In my code I've been matching against the full string returned first with checks like `if err:match("^ENOENT:") then...`

rphillips:
ok. previously we would get the raw error code in luvit... i do sorta like the string.

PR #122, opened 2015-03-18
cleanup handle on error

rphillips:


src/handle.c
```diff
@@ -70,7 +70,7 @@ static void luv_close_cb(uv_handle_t* handle) {
 static int luv_close(lua_State* L) {
   uv_handle_t* handle = luv_check_handle(L, 1);
   if (uv_is_closing(handle)) {
-    luaL_error(L, "handle %p is already closing", handle);
+    return luaL_error(L, "handle %p is already closing", handle);
```
creationix: Does this suppress a compiler warning?  `luaL_error` never returns since it does a `longjmp` as part of the exception process.

creationix:
+1

src/handle.c
rphillips: fixed

creationix: I didn't mind it with the return, I was just curious.

Keep it whichever way looks better to you.  I imagine it might affect GCC's optimizer a little when doing path analysis.

PR #123, opened 2015-03-19
luarocks support.

xpol:
Add support for luarocks.

To release:

1. bump version number in rockspec filename and file contents.
1. tag `v1.4.2` in github
2. upload rockspec to  https://rocks.moonscript.org

See [Creating a rock](/keplerproject/luarocks/wiki/Creating-a-rock)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,9 +3,14 @@ cmake_minimum_required(VERSION 2.8)
 project (luv)
 
 include(deps/uv.cmake)
-include(deps/luajit.cmake)
 
-include_directories(deps/luajit/src)
+if (LUV_LUA_INCDIR)
+  include_directories(${LUV_LUA_INCDIR})
+  link_directories(${LUV_LUA_LIBDIR})
+else(LUV_LUA_INCDIR)
+  include(deps/luajit.cmake)
+  include_directories(deps/luajit/src)
+endif(LUV_LUA_INCDIR)
 
 add_library (luv MODULE src/luv.c)
 
@@ -31,3 +36,6 @@ else()
   target_link_libraries(luv uv)
 endif()
 
+if (LUV_INSTALL_PREFIX)
+  install(TARGETS luv DESTINATION "${LUV_INSTALL_PREFIX}/lib")
+endif(LUV_INSTALL_PREFIX)
--- a/luv-1.4.2-1.rockspec
+++ b/luv-1.4.2-1.rockspec
@@ -1,0 +1,33 @@
+package = "luv"
+version = "1.4.2-1"
+source = {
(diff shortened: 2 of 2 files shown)


creationix:
Thanks.  Two nits:

 - The License is actually Apache 2.0 to match Luvit
 - Can you squash your commits into one?

xpol:
Done. :smile:

creationix:
Does this look right? http://rocks.moonscript.org/modules/creationix/luv

creationix:
It doesn't seem to work, cmake is missing files.  Maybe luarocks doesn't include git submodules?

xpol:
Oh, yes.
The git submodule maybe not supported by luarocks. (I will discuss this with luarocks project.)

Is it possible that we switch from submodules to subtree ?
subtree has many advantages over submodules.

xpol:
Added a issue here: keplerproject/luarocks#334 .

creationix:
I'm not going to switch to subtree.  I know the tradeoffs and prefer submodules for this project.

I did try manually creating the .src.rock file with the libuv submodule included (the packer didn't include it automatically so I added it to the zip), but when I tried to install, it ignored my uploaded zip file and did a non-recursive clone anyway.

Maybe we should add a cmake command to init and update the submodule if deps/libuv/src/uv.h is missing?

xpol:
> Maybe we should add a cmake command to init and update the submodule if deps/libuv/src/uv.h is missing?
Good idea! :+1:

creationix:
Doesn't help.  I read somewhere that luarocks deletes the .git folder after cloning.  If this is true, the submodule update command I added will have no effect.

https://github.com/luvit/luv/commit/d0664a69756296a6d0bcdaf3f909e8fb65197ebe

creationix:
Yep https://github.com/keplerproject/luarocks/blob/master/src/luarocks/fetch/git.lua#L81

joerg-krause:
How about providing an uploaded release tarball instead of a tag like it's done for luvi? Note, I don't mean the automatically generated tarball by github, but an uploaded one. This has some advantages for package maintainers and build systems. I would also suggest to use a proper release tarball name in case of luvi: <name>-<tag>, e.g. luvi-v1.0.1.tar.gz, luv-v1.4.2.tar.gz.

Using a release tarball there is no need for any git commands in a Makefile. A github user can still follow the "Building From Source" section to get a working copy, a package maintainer or build system can download the release tarball.

creationix:
Ok, I manually created a github release with a hand-crafted tarball.  This seems to work now.  Try with luv 1.4.2-2 on luarocks.

xpol:
Works fine! @creationix Thank you!

Issue #127, opened 2015-03-26
Make all C -> Lua calls protected and allow user error handlers in uv.run()


creationix:
This task will fix #99.

The idea is to protect all calls into Lua from C using pcall.  If there is an error, it will be handled by the current global error handlers.  If there is no current error handler then we'll throw an unprotected lua error (most likely this will crash the process).

Since calls to `uv.run()` are blocking, then "nested" calls will be all "stacked" in the current C stack.  Therefore users can have nested error handlers.  A unit test framework may wish to handle all global errors, but only for the duration of it's test unit.  It can call and block on `uv.run()` with a custom error handlers that pertains to that one test.  When the test finishes and `uv.run()` returns, then the old error handler will be restored.

The new signature will be `uv.run(mode, onError)`.

rphillips:
Perhaps make the errors a ring. Just to make sure it isn't unbounded.

creationix:
That's not a problem, errors thrown by error handlers won't feed back into themselves.

creationix:
My implementation idea is to add a couple global variables per luv instance.  One will hold a reference to the current error handler function.  The other will be a temporary place to hold the error value when an error happens.

The global error handler on the C side will simply store the error in the global place and unblock `uv_run()` with `uv_stop()`.  Then the `luv_run()` wrapper will check the global error slot for a possible error.

creationix:
WIP at #129

creationix:
I decided this is a very deep rabbit-hole with little benifit.  I instead made a simple change to `lua_pcall` all callback event sources and hard-error if there is an uncaught exception.  All existing luv, luvi, and luvit unit tests still pass with this change since we almost always wrap our code in xpcall to get stack traces.

This change will simply enforce the current de-facto behavior and let you know if you forget to protect your event sources.

creationix:
Implemented in https://github.com/luvit/luv/commit/7526c1d9955f085df9eb1c25fcc31c0cfe67e583

neoxic:
Probably should be reopened if you allow my humble five cents...

[...]

PR #129, opened 2015-03-26
WIP: global error handling

creationix:


creationix:
Not a good idea, see rational and new patch in #127

PR #132, opened 2015-04-02
fix bug in fs_event_start callback, file fs_event.c

zhaozg:
according to uvbook, http://nikhilm.github.io/uvbook/filesystem.html#file-change-events

```lua
local uv = require('luv')

if #arg==0 then
    print(string.format("Usage: %s <command> <file1> [file2 ...]",arg[0]));
    return
end

for i=1,#arg do
    local fse = uv.new_fs_event()
    assert(uv.fs_event_start(fse,arg[i],{
        --"watch_entry"=true,"stat"=true,
        recursive=true
    },function (err,fname,status)
        if(err) then
            print("Error "..err)
        else
            print(string.format('Change detected in %s',
                uv.fs_event_getpath(fse)))
            for k,v in pairs(status) do
                print(k,v)
            end                
            print('file changed:'..(fname and fname or ''))
        end
    end))
    
end

uv.run('default')
uv.loop_close()

```

### DIFF
--- a/src/fs_event.c
+++ b/src/fs_event.c
@@ -45,17 +45,14 @@ static void luv_fs_event_cb(uv_fs_event_t* handle, const char* filename, int eve
 
   // events
   lua_newtable(L);
-  if (events & UV_FS_EVENT_WATCH_ENTRY) {
+  if (events & UV_RENAME) {
     lua_pushboolean(L, 1);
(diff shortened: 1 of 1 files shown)


zhaozg:
@creationix  @rphillips

rphillips:
+1 thanks

PR #133, opened 2015-04-03
add basic thread support

zhaozg:
Any advice, will with more feature,
come from uvbook, thread-create.lua
```lua
local uv = require('luv')

local hare_id = uv.new_thread()
local tortoise_id = uv.new_thread()
local step = 10

uv.thread_create(hare_id,function(step) 
    local ffi = require'ffi'

    local sleep 
    if ffi.os=='Windows' then
        ffi.cdef "void Sleep(int ms);"
        sleep = ffi.C.Sleep
    else 
        ffi.cdef "unsigned int usleep(unsigned int seconds);"
        sleep = ffi.C.usleep
    end
    while (step>0) do 
        step = step - 1
        sleep(1000)
        print("Hare ran another step")
    end
    print("Hare done running!")
end, step,true,'abcd','false')

uv.thread_create(tortoise_id,function(step)
    while (step>0) do 
        step = step - 0.5
        print("Tortoise ran another step")
    end
    print("Tortoise done running!")
end,step,'abcd','false')

print(hare_id==hare_id,uv.thread_equal(hare_id,hare_id))
print(tortoise_id==hare_id,uv.thread_equal(tortoise_id,hare_id))

uv.thread_join(hare_id)
uv.thread_join(tortoise_id)
```

### DIFF
--- a/src/luv.c
+++ b/src/luv.c
@@ -40,7 +40,7 @@
 #include "fs.c"
 // #include "work.c"
 #include "dns.c"
-// #include "thread.c"
+#include "thread.c"
 #include "misc.c"
 #include "constants.c"
 
@@ -241,6 +241,13 @@ static const luaL_Reg luv_functions[] = {
   {"uptime", luv_uptime},
   {"version", luv_version},
   {"version_string", luv_version_string},
+
+  // thread.c
+  {"new_thread", luv_new_thread},
+  {"thread_create", luv_thread_create}, 
+  {"thread_equal", luv_thread_equal},
+  {"thread_self", luv_thread_self},
+  {"thread_join", luv_thread_join},
 
   {NULL, NULL}
 };
@@ -465,6 +472,8 @@ LUALIB_API int luaopen_luv (lua_State *L) {
   luaL_newlib(L, luv_functions);
   luv_constants(L);
   lua_setfield(L, -2, "constants");
+
+  luv_thread_init(L);
 
   return 1;
 }
--- a/src/luv.c
+++ b/src/luv.c
@@ -248,7 +248,8 @@ static const luaL_Reg luv_functions[] = {
   {"thread_equal", luv_thread_equal},
   {"thread_self", luv_thread_self},
   {"thread_join", luv_thread_join},
-
+  {"sleep", luv_thread_sleep},
+  
   // work.c
   {"new_work", luv_new_work},
   {"queue_work", luv_queue_work},
--- a/src/luv.c
+++ b/src/luv.c
@@ -38,9 +38,9 @@
 #include "fs_event.c"
 #include "fs_poll.c"
 #include "fs.c"
-// #include "work.c"
 #include "dns.c"
 #include "thread.c"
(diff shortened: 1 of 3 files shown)


creationix:
Thanks for getting this started.  I'll look into it.

creationix:
I'm a little concerned about threads having access to side-effects like `print`, especially if that's via libuv.

I think we could make something like web-workers quite safely.  We could even use `uv_async` to send messages between threads.

src/luv.c
```diff
@@ -242,6 +242,13 @@ static const luaL_Reg luv_functions[] = {
   {"version", luv_version},
   {"version_string", luv_version_string},
 
+  // thread.c
+  {"new_thread", luv_new_thread},
+  {"thread_create", luv_thread_create}, 
+  {"thread_equal", luv_thread_equal},
+  {"thread_self", luv_thread_self},
+  {"thread_join", luv_thread_join},
+
```
creationix: We should also add the method form for `:create`, `:equal`, and `:join` in the metatable for threads so you can do:

```lua
uv.new_thread():create(function () 
  -- ...
end)
```

Also, is there a reason to have `uv.thread_create` be apart from `uv.new_thread`?

zhaozg:
Yes,thread and work need more think and good design,it it long term work,now just begin.

src/luv.c
zhaozg: New only create handle,and create real start thread,

zhaozg: New only create handle,and create real start thread,please don't cost more time on it,because will  change lot in afew days.

zhaozg:
It hard work to make lua script  as work_cb and after_work_cb. serial callback code everytime will a lots of CPU time

zhaozg:
Hi @creationix,  When I read http://nikhilm.github.io/uvbook/threads.html#inter-thread-communication, I realize that 
uv_async_t is design for ITC(inter-thread-communication) , main loop thread create it and share it with thread or threadpool.
But according https://github.com/luvit/luv/blob/master/src/async.c#L31-L52 and https://github.com/luvit/luv/blob/master/docs.md#uv_async_t--async-handle
, after uv.new_async create async object,we have no method to pass to sub or others threads.

When try to code http://nikhilm.github.io/uvbook/threads.html#inter-thread-communication with luv, it cause async.c design or find a way share userdata.

In my options, async should behind in thread.c and work.c, maybe will broken freezed code.

zhaozg:
work demo
```lua
local uv = require('luv')

local ctx = uv.new_work(
    function(n) --work,in threadpool
        local uv = require('luv')
        local t = uv.thread_self()
        print(t)
        uv.sleep(100)
        return n*n,n 
    end, 
    function(r,n) print(string.format('%d => %d',n,r)) end    --after work, in loop thread
)
uv.queue_work(ctx,2)
uv.queue_work(ctx,4)
uv.queue_work(ctx,4)
uv.queue_work(ctx,4)
uv.queue_work(ctx,4)

uv.run('default')
uv.loop_close()
```

PR #135, opened 2015-04-05
add basic thread support

zhaozg:
include #133


rphillips:
This is awesome. A few test cases would be extremely nice.

zhaozg:
some example https://github.com/zhaozg/luv/blob/master/examples/uvbook/queue-work.lua and https://github.com/zhaozg/luv/blob/master/examples/uvbook/thread-create.lua

rphillips:
+1

creationix:
@zhaozg is this safe to merge or did you want to work on it some more?

zhaozg:
Please merge, It enough support to thread and simple threadpool.

imzyxwvu:
I do not think it is a good idea to call string.dump instead of lua_dump.

zhaozg:
>I do not think it is a good idea to call string.dump instead of lua_dump.

Accept, I will update

PR #139, opened 2015-04-14
Remove submodule handling from CMakeLists.txt; Minor fixes

joerg-krause:
Dependency management like checkout or cloning submodules (dependencies) is not the task of the project build files, but of the builder. Of course the builder needs instructions about how to build and what dependencies are needed. This fits well in the way package manager work. For example in Buildroot or Arch Linux you define that the package luv depends on the packages libuv and luajit. This way the package manager takes care that the dependencies are build before luv.


creationix:
Thanks, this looks much cleaner now.

PR #142, opened 2015-04-17
fix build warning on windows

zhaozg:


rphillips:
thanks!

src/thread.c
```diff
@@ -17,6 +17,7 @@
 #include "luv.h"
 
 #include "lthreadpool.h"
+#include <lualib.h>
```
creationix: Should we move this include to luv.h alongside the lua.h and luaxlib.h includes?

rphillips: yeah, that is a good idea

PR #145, opened 2015-04-22
Add support for building luv as static library

joerg-krause:
This commit adds on option `BUILD_MODULE` which defaults to `ON`. The user can
set this option to `OFF` to build luv as static library. If `BUILD_MODULE` is
set to `OFF` and `BUILD_SHARED_LIBS` is set to `ON` a shared library will be
build instead.


joerg-krause:
Thanks for merging!

Issue #147, opened 2015-04-29
Is there any easy way to using luv/luvi/luvit with Lua5.3?


starwing:
Hi,

this projects is very interesting :) I have built it from source without any issue. and the bundle idea is GREAT!!

As I use PUC Lua for my daily work. I want to get some information about it:
  - can I (personally) change it's LuaJIT to PUC Lua 5.3 easily?
  - If I do that, can I put some so/dlls into bundle? and how luvit work in this way?
  - or, is there any other way to use extenal dll libraries for Lua?

I built all my modules againest Lua5.3, So change Lua in luvi seems easiler. Just wonder whether a way exits.

thanks!

imzyxwvu:
Only the binding for libuv could work with PUC Lua. luvit is highly based on luajit and lua 5.1. For now there is no way to bundle Lua C libraries. The best way to call C code is the luajit C ffi.

creationix:
There are a few places in luvi and luvit that assume luajit semantics.  The hardest one would be the ffi interface (unless there is a C replacement like there is for the bit.* apis).

Luvit's require system can load binary addons from the bundle and lit supports publishing binary libraries to the public database.

When you try to require a .so/.dll from the bundle, luvi will extract it to a temporary file, dlopen it and delete the temporary file.  This is the same technique java jar files use so it's pretty well tested.

https://github.com/luvit/lit/wiki/Publishing-Compiled-Code

creationix:
I will say that the luv (libuv bindings for lua) are engine agnostic and should work with PUC lua 5.3, if not, file an issue and I'll fix it, it's a supported use case.

https://github.com/luvit/luv

Also the "luvit" project is just a set of lua libraries on top of luvi.  You might be able to reuse them with minimal modification using luv and stock lua (basically skipping the whole luvi layer).  You'd have to either create replacements for the luvi APIs or modify the luvit libraries that use luvi APIs.

creationix:
See also luvit/luvit#682

starwing:
@creationix Thank you for your detail explains :)

I have found some imcompatible with Lua5.3, and make a pull request about that.
btw, I tweak the build script so that PUC Lua is included into source, and we can choose build luv against with PUC Lua in cmake options.

The news about support binary module is great! I will try to build some of my modules (mainly about charset) to luvit.

thanks :)

creationix:
[...]

PR #148, opened 2015-04-30
Make luv compatible with Lua5.3

starwing:
- add function luv_check_buf() to avoid warnings in MinGW (convert DWORD* to size_t*)
- avoid MinGW warnings that defined macros (S_IS* macros)
- change luaL_checkint to luaL_checkinteger
- use lua_getglobal() instead LUA_GLOBALSINDEX

some improves againest #147

### DIFF
--- a/src/luv.h
+++ b/src/luv.h
@@ -77,8 +77,6 @@ LUALIB_API int luaopen_luv (lua_State *L);
 #include "lhandle.h"
 #include "lreq.h"
 
-// From luv.c
-
 // From stream.c
 static uv_stream_t* luv_check_stream(lua_State* L, int index);
 static void luv_alloc_cb(uv_handle_t* handle, size_t suggested_size, uv_buf_t* buf);
--- a/src/luv.h
+++ b/src/luv.h
@@ -29,11 +29,13 @@
 # include <fcntl.h>
 # include <sys/types.h>
 # include <sys/stat.h>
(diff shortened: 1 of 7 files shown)


src/luv.h
```diff
@@ -29,11 +29,13 @@
 # include <fcntl.h>
 # include <sys/types.h>
 # include <sys/stat.h>
-# define S_ISREG(x)  (((x) & _S_IFMT) == _S_IFREG)
-# define S_ISDIR(x)  (((x) & _S_IFMT) == _S_IFDIR)
-# define S_ISFIFO(x) (((x) & _S_IFMT) == _S_IFIFO)
-# define S_ISCHR(x)  (((x) & _S_IFMT) == _S_IFCHR)
-# define S_ISBLK(x)  0
+# ifndef __MINGW32__
```
creationix: What's this for?  Are these already defined to something else in mingw32?

starwing: Yes, In MinGW's sys/stat.h, these macros are all defined.

src/luv.h
```diff
@@ -75,9 +77,13 @@ LUALIB_API int luaopen_luv (lua_State *L);
 #include "lhandle.h"
 #include "lreq.h"
 
+// From luv.c
+
```
creationix: Did you mean to add these comments?

starwing: Sorry for that :( that was remaining from previous refator, I will remove that.

PR #149, opened 2015-04-30
Add Lua5.3 to tree and tweak the build script.

starwing:
now support:
- PUC lua can be built with luv.
- now can build dll library against lua/luajit (lua53.dll/lua51.dll)

so we support all combines below:
- static/dynamic luv library
- static/dynamic/system Lua library
- PUC Lua/LuaJIT implement


src/thread.c
```diff
@@ -129,7 +129,11 @@ static const char* luv_thread_dumped(lua_State* L, int idx, size_t *l) {
   } else {
     const char* buff;
     luaL_checktype(L, idx, LUA_TFUNCTION);
+#if LUA_VERSION_NUM >= 502
+    lua_getglobal(L, "string");
```
creationix: How about we add this as a macro like the other lua 5.3 macros in luv.h?

starwing: Good catch! Yes, I have found in Lua5.1 source, it already have lua_[gs]etglobal. So I will use that :)

creationix:
@joerg-krause could you look over this PR and make sure it doesn't break any of your stuff?

starwing:
Some notes: I noticed that all projects used luv all generate "luajit.sln" for Visual Studio Solution name. That's because one line "project(luajit C ASM)" in deps/luajit.cmake, because *.cmake is use include() to load, not add_subdirectory(). So I removed this line, and add ASM to CMakeLists.txt. Now the solution name is "luv", as expected.

creationix:
@starwing good catch, I was wondering how to fix that for windows.

starwing:
@creationix Now rebased above to @joerg-krause 's commit. test on Windows with: Lua Static/ LuaJIT Static/Lua Dynamic/LuaJIT Dynamic. I don't have *nix environment so I can't test on that :( Maybe someone can help with it...

CMakeLists.txt
```diff
@@ -10,7 +10,26 @@ set(LUV_VERSION ${LUV_VERSION_MAJOR}.${LUV_VERSION_MINOR}.${LUV_VERSION_PATCH})
...
+    CACHE STRING "Link to LuaJIT or PUC Lua" FORCE)
+  set_property(CACHE WITH_LUA_ENGINE
+    PROPERTY STRINGS  "Lua;LuaJIT")
+endif (NOT WITH_LUA_ENGINE)
+
+if (NOT LUA_BUILD_TYPE)
+  set(LUA_BUILD_TYPE "Static"
+    CACHE STRING "Build Lua/LuaJIT as static, dynamic libary, or use system one" FORCE)
+  set_property(CACHE LUA_BUILD_TYPE
+    PROPERTY STRINGS  "Static;Dynmaic;System")
```
rphillips: typo on dynamic

starwing: Sorry for that :(

joerg-krause:
@creationix Sorry, I did not noticed your comment. I will have a look at it.

starwing:
@joerg-krause Thank you! but it fails on Linux, and I don't have Linux environment, So I need help :)

joerg-krause:
@starwing The Travis CI build error is caused by a bug in libuv. The issue is fixed in libuv 1.5.0. So please fix the error mentioned in my previous comment rebase to the latest commit with libuv 1.5.0 and try again.

joerg-krause:
I am not so sure about the build error. It seems to happen randomly and even happened with the latest commit.

starwing:
@joerg-krause Thank you for reply, I will fix that :)

I will look into the build error if I have time.

starwing:
@joerg-krause Thank you for reply, I will fix that :)

I will look into the build error if I have time.

PR #151, opened 2015-05-02
Rename target library name from lua51 to luajit-5.1

joerg-krause:
luajit-5.1 is the standard name for the LuaJIT library. Use this name to
prevent issues when linking against a shared LuaJIT library.


creationix:
@joerg-krause I'm trying to make a luvi release, but on machines that have luajit installed as a system library, the build system is using that instead of using the embedded luajit.  I really need to use the embedded luajit in luv since we pass in custom config options to luajit.  Do you know how to help?

joerg-krause:
@creationix Which config options?

creationix:
@joerg-krause Stuff like enabling lua52 compat which distribution builds of luajit never seem to enable.

joerg-krause:
A clean solution would be to check if the found luajit library has the required additional features built-in. If not, the build system may throw an error and a hint to build luv with `WITH_SHARED_LUAJIT=OFF`.

creationix:
@joerg-krause no, my problem is that I'm using WITH_SHARED_LUAJIT=OFF and it still uses the system luajit if there is one, but not if there is not.

Luvi itself doesn't require the custom lua features.

joerg-krause:
@creationix I cannot reproduce this on my system. If I build luv with `WITH_SHARED_LUAJIT=OFF make clean all` it always builds the embedded luajit. It's the same for luvi with `WITH_SHARED_LUAJIT=OFF make clean tiny && make`.

Can you post a build log, please.

creationix:
Interesting.  I'm building luvi which uses luv. Perhaps it's not passing the right options down to luv?  I would expect it to at lease use the default value though.

creationix:
Ok, my submodule for luv wasn't up to date and luvi, instead of throwing an error, just used the system libluajit  When I disabled the system libluajit I saw the error more clearly.

joerg-krause:
@creationix Yes, submodules are sometimes tricky. You can use `git config status.submodulesummary 1` so Git will also show you a short summary of changes to the submodules.

creationix:
Thanks for the tip!  And sorry for the trouble.

PR #157, opened 2015-06-02
add new_socket_poll method

### LINKED ISSUE #156: ENOTSOCK: socket operation on non-socket
I run this code and return "ENOTSOCK: socket operation on non-socket"

````lua  
local tcp = require('coro-tcp')
local uv = require('uv')

local r, w, socket = tcp.connect(ip, port)
    if nil == r then
    return nil, w
end
local sockfd = socket:fileno()
local poll, e = uv.new_poll(sockfd)
if e then return e end
````

ENOTSOCK: socket operation on non-socket

### PR DESCRIPTION
**Pull request by runner-mei**
this request is fix #156

### DIFF
--- a/src/poll.c
+++ b/src/poll.c
@@ -34,6 +34,18 @@ static int luv_new_poll(lua_State* L) {
   return 1;
 }
 
+static int luv_new_socket_poll(lua_State* L) {
+  int fd = luaL_checkinteger(L, 1);
+  uv_poll_t* handle = lua_newuserdata(L, sizeof(*handle));
+  int ret = uv_poll_init_socket(luv_loop(L), handle, fd);
+  if (ret < 0) {
+    lua_pop(L, 1);
+    return luv_error(L, ret);
+  }
+  handle->data = luv_setup_handle(L);
+  return 1;
+}
+
 // These are the same order as uv_run_mode which also starts at 0
 static const char *const luv_pollevents[] = {
   "r", "w", "rw", NULL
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/poll.c [hunk 1]
```diff
@@ -34,6 +34,18 @@ static int luv_new_poll(lua_State* L) {
   return 1;
 }
 
+static int luv_new_socket_poll(lua_State* L) {
+  int fd = luaL_checkinteger(L, 1);
```
rphillips: This needs a tweak for windows. The fd will be a SOCKET handle.

creationix: Don't these survive being stored in ints as long as `sizeof(int) == sizeof(uintptr_t)`?  We use int for file descriptors in the fs apis.

rphillips: hmm. yeah, nice call. +1

creationix: Well, what I said is wrong, ints are usually 32-bit, but pointers are 64-bit on x86_64 and amd64.  But my comment about using ints for FDs is still true.

rphillips: ack. yeah, this is a hard one.

runner-mei: this bug isn't cause by convert socket handle to int, it is invoke uv__get_osfhandle in the uv_poll_init

````lua
int uv_poll_init(uv_loop_t* loop, uv_poll_t* handle, int fd) {
  return uv_poll_init_socket(loop, handle, (SOCKET) uv__get_osfhandle(fd));
}
````

PR #160, opened 2015-06-25
don't free the buff if it is null.

DBarney:
http://docs.libuv.org/en/latest/stream.html#c.uv_read_start says that but.base can be set to 0 in certain cases. This accounts for that, and only frees the bf.base if it has been set to a value.


justinmk:
It's harmless to send NULL to free().

DBarney:
yep you are right.

PR #164, opened 2015-07-17
bump luajit 21

rphillips:
trying out a luajit 2.1 bump

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,10 +1,16 @@
 configuration: Release
 
+environment:
+  matrix:
+    - LUV_ARCH: amd64
+    - LUV_ARCH: ia32
+
 matrix:
   fast_finish: true
 
 build_script:
  - git submodule update --init
+ - msvcbuild.bat %LUV_ARCH%
  - msvcbuild.bat
 
 test_script:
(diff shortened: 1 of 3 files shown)


joerg-krause:
Is there any need to switch to (unstable) luajit 2.1? I would rather see luv to depend on a stable version.

creationix:
I want to keep the default build on stable luajit, but I'm all for adding the option to use luajit 2.1.  I wonder how we can support both at the same time.

DBarney:
I haven't had any issues with 2.1, in fact its nice to be able to use the built in profiler and generate flame graphs to optimize hot code paths.

The 2.1 branch is mostly stable, I saw on the mailing list a few weeks ago that they are considering making a 2.1 release soon, just have to finish off a few last features.

maybe instead of making 2.1 the default, it could just be a build option?

joerg-krause:
If it's only about performance I would stick to luajit 2.0.4 until luajit 2.1 is finally released.

>  I wonder how we can support both at the same time.

How about a new branch for luajit 2.1?

rphillips:
I cherry-picked my 32bit tests commit to a new PR.

PR #169, opened 2015-07-28
process: cleanup the lua handle in the close callback

rphillips:
* The close callback happens after the endgames, so it is safe to
  destroy


creationix:
+1

PR #170, opened 2015-08-02
Set correct installation path for luv module on non-WIN32 targets

joerg-krause:
The search path for loading binary modules with the require function is
`LUA_ROOT "lib/lua/" LUA_VERSION_MAJOR "." LUA_VERSION_MINOR` on non-WIN32
targets.

Let us use this default path for installing the luv module. This allows the user to use

```lua
local uv = require('luv')
```

from every directory path.


creationix:
I've never used this use case, but it sounds like a great change. +1

rphillips:
+1

PR #173, opened 2015-08-12
fix(errors): improve error message in tcp.c and udp.c

rphillips:


creationix:
The argerror doesn't seem to be providing much, can we just switch to luaL_error which has printf functionality built in?

rphillips:
sure. I'll change the PR

PR #180, opened 2015-08-25
cmake: add thread library

rphillips:


zhaozg:
this is not need, the problem solved int #181

rphillips:
cool. thanks

PR #189, opened 2015-11-02
fix two problems in windows

mythay:
1. export libuv symbols when linking without shared libuv
2. fix linking lua lib name without standard name in windows (lua for windows)


rphillips:
+1 Thank you.

CMakeLists.txt
```diff
@@ -136,7 +139,11 @@ if(WIN32)
   if (USE_LUAJIT)
     target_link_libraries(luv uv luajit-5.1)
   else (USE_LUAJIT)
-    target_link_libraries(luv uv lualib)
+    if (LUA_BUILD_TYPE STREQUAL System)
+        target_link_libraries(luv uv ${LUA_LIBRARIES})
+    else (LUA_BUILD_TYPE STREQUAL System)
```
creationix: This looks like the second case will never be true?  Was this s typo?

PR #200, opened 2015-12-12
Add missed netmask filed in network interface

zhaozg:
This patch will add missed netmask field in network interface.


zhaozg:
http://docs.libuv.org/en/v1.x/misc.html#c.uv_interface_address_t

rphillips:
thanks. lgtm +1

PR #206, opened 2015-12-31
change luajit submodule url to https://github.com/LuaJIT/LuaJIT.git

zhaozg:
fix travis fail, because luajit can't access.


rphillips:
+1

creationix:
This should work.  The github mirror is just a mirror, but I believe it will stay up to date since lots of development happens there now.

PR #210, opened 2016-01-26
Fix AppVeyor on Pro environment

tbastos:
Solution as suggested in http://help.appveyor.com/discussions/problems/4031-same-code-in-two-repositories-build-passes-in-one-repo-but-fails-in-the-other


creationix:
Awesome work!

PR #211, opened 2016-01-27
Change fs_scandir_next() to return two values instead of a table

### LINKED ISSUE #201: Function fs_scandir_next creates tables unnecessarily
I think `fs_scandir_next` is being unnecessarily wasteful here: https://github.com/luvit/luv/blob/master/src/fs.c#L406-L436

Instead of creating a table at each call it could simply return 2 values (name and type). For programs that scan lots of big dirs, this is going to be much faster, and more importantly won't create a lot of garbage and put a lot of pressure on the GC.

The resulting code is also simpler to use. A table only adds complexity here, IMHO.

tbastos:
Changes:
- Closes #201 (Change `fs_scandir_next()` to return two values instead of a table)
- Increase interval tolerance in test "timer - timeout with interval" in order to avoid random fails on AppVeyor and slow CI servers. Increased from 50ms to 100ms.

### DIFF
--- a/src/fs.c
+++ b/src/fs.c
@@ -416,24 +416,21 @@ static int luv_fs_scandir_next(lua_State* L) {
     return 0;
   }
   if (ret < 0) return luv_error(L, ret);
-  lua_createtable(L, 0, 2);
   lua_pushstring(L, ent.name);
-  lua_setfield(L, -2, "name");
+  if (ent.type == UV_DIRENT_UNKNOWN) return 1;
   switch (ent.type) {
-    case UV_DIRENT_UNKNOWN: type = NULL; break;
-    case UV_DIRENT_FILE: type = "file"; break;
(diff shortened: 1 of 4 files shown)


tbastos:
PS: do you mind uploading a new luarocks release after this? I'll need it for my project and it can't be done with a PR, unfortunately. Thanks!

creationix:
It says there are conflicts.  Can you rebase the PR with master?

tbastos:
Conflict fixed. Please don't forget the new luarocks release ;)

creationix:
Ok, I think it's published. http://luarocks.org/modules/creationix/luv/1.8.0-1

tbastos:
Thanks! My AppVeyor is almost functional again :)

I thought you'd bump the version number since there were breaking API changes, but it doesn't really matter for me.

creationix:
I would, but the version number matches the upstream libuv version which we didn't change.

PR #213, opened 2016-02-04
Fix incorrect behaviour on Windows related to stdio pipes being read-write

tbastos:
Unfortunately if we always always create the stdio pipes as R+W (instead of only R for stdin, and only W for stdout and stderr) child processes on windows start to behave incorrectly. For example, if you spawn the 'cat' program, write to it and then shutdown its stdin, the correct behaviour is for the child to receive EOF and exit. But if stdin is RW the child never exits on Windows, and you have to kill the process to force it to exit. Changing the stdin pipe to only `UV_CREATE_PIPE|UV_READABLE_PIPE` fixes the problem.

The current solution assumes only the standard stdio fd's (0-2) are used. A more flexible solution may be in order, but would require an API change.

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -111,7 +111,8 @@ static int luv_spawn(lua_State* L) {
         uv_stream_t* stream = luv_check_stream(L, -1);
         int err = uv_fileno((uv_handle_t*)stream, &fd);
         if (err == UV_EINVAL || err == UV_EBADF) {
-          options.stdio[i].flags = UV_CREATE_PIPE | UV_READABLE_PIPE | UV_WRITABLE_PIPE;
+          options.stdio[i].flags =
+              UV_CREATE_PIPE | (i == 0 ? UV_READABLE_PIPE : UV_WRITABLE_PIPE);
         }
         else {
           options.stdio[i].flags = UV_INHERIT_STREAM;
@@ -257,4 +258,3 @@ static int luv_kill(lua_State* L) {
   lua_pushinteger(L, ret);
   return 1;
 }
-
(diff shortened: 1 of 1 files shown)


src/process.c
```diff
@@ -111,7 +111,8 @@ static int luv_spawn(lua_State* L) {
         uv_stream_t* stream = luv_check_stream(L, -1);
         int err = uv_fileno((uv_handle_t*)stream, &fd);
         if (err == UV_EINVAL || err == UV_EBADF) {
-          options.stdio[i].flags = UV_CREATE_PIPE | UV_READABLE_PIPE | UV_WRITABLE_PIPE;
+          options.stdio[i].flags =
+              UV_CREATE_PIPE | (i == 0 ? UV_READABLE_PIPE : UV_WRITABLE_PIPE);
```
creationix: This is great, but I'm unsure it's something we can assume.
We should probably use read/write for fds above 2 right? 
How are higher FDs typically used?  I've used them to pass TCP FDs from master load balancer to child worker processes.

tbastos:
> We should probably use read/write for fds above 2 right?
Cool, that should cover all use cases... I'll send another commit shortly.

tbastos:
If everything is alright, do you mind uploading a new LuaRocks release? I'll need it to fix my AppVeyor. Thanks!

creationix:
Thanks for fixing things on windows.  I really want luv to be stable and supported on windows, but I don't have the time or resources to do that properly.  Your help is greatly appreciated.  I'll publish a new rock shortly.

creationix:
Published to github and luarocks.

- https://github.com/luvit/luv/releases/tag/1.8.0-2
- http://luarocks.org/modules/creationix/luv/1.8.0-2

Let me know if this works.

tbastos:
Thank you, everything is working now! I'm happy to help, thanks for being receptive :+1: 

I know Windows is productivity-sucking black hole if you're used to developing on Linux/OSX. I'm on OSX but since so many people still use Windows (unfortunately), I'm willing to do the dirty work

PR #214, opened 2016-02-06
bump to luajit 2.1

rphillips:
* 64bit integer conversions do not work on Windows with 2.0.4

`29913653248` is converted to `0xf6fe2000`. Note: The low 32bits are correct, but the high 32 bits is missing. The correct answer is: `0x6f6fe2000`.

`32207650816` is converted to `0x7fb9c000`. The correct answer is `0x77fb9c000`.

Update: Luajit 2.1 does fix the issue on Windows.


creationix:
+1

PR #221, opened 2016-02-25
fix wrong lua_one.c location in non Windows systems.

starwing:


creationix:
Doesn't this disable the `WITH_AMALG` option?

creationix:
Nevermind, I needed to read more context.  Looks good.

PR #224, opened 2016-03-08
Use luarocks native build on rockspec

### LINKED ISSUE #223: Can't create 32-bit build
neovim build system downloads a bunch of dependencies, including a number of luarocks modules required during the build process such as lpeg and a client library based on libuv that now uses luv as the backend(instead of a simple libuv binding module).

The problem is that luv.so is always built as a 64-bit library, even when doing 32-bit builds with both CFLAGS and LDFLAGS set to -m32. Normally we can force luarocks to install 32-bit modules by passing `gcc -m32` as `CC` and `LD`, but this seems to be failing with luv

I also tried to build from luv source tree:

```
mkdir build
cd build
CC='gcc -m32' LD='gcc -m32' cmake ..
```

which fails with:

```
-- The C compiler identification is GNU 5.2.1
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/gcc
-- Check for working C compiler: /usr/bin/gcc
-- Check for working C compiler: /usr/bin/gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of void*
-- Check size of void* - done
-- Looking for dlopen
-- Looking for dlopen - not found
CMake Error at deps/luajit.cmake:84 (MESSAGE):
  Cannot compile a useful lua.

  Function dlopen() seems not to be supported on your platform.

  Apparently you are not on a Windows platform as well.

  So lua has no way to deal with shared libraries!
Call Stack (most recent call first):
  CMakeLists.txt:106 (include)
```

tarruda:
This should make luv more compatible with luarocks echosystem:

- Doesn't depend on cmake(which may not be available). luv is very simple to
  compile, and cmake is used more to setup dependencies, but this is already
  taken care of by luarocks.
- Allows the user to specify custom libuv prefixes with LIBUV_DIR(which is the
  idiomatic way of specifying dependencies location in luarocks)
- Respects CC

Close #223

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,13 @@ script:
   - source .ci/setenv_lua.sh
   - luarocks make
   - test $PWD = `lua -e "print(require'luv'.cwd())"`
+  - luarocks remove luv
+  # Test the alternate rockspec
+  - mkdir build/lib
+  - cp build/libuv.a build/lib
+  - cp -a deps/libuv/include build
+  - luarocks make rockspecs/$(ls rockspecs) LIBUV_DIR=build
+  - test $PWD = `lua -e "print(require'luv'.cwd())"`
 
 notifications:
   email: true
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -36,6 +36,13 @@ build_script:
   - luarocks make
   - ps: if("$(Get-Location)" -eq $(lua -e "print(require'luv'.cwd())")) { "LuaRocks test OK" } else { "LuaRocks test failed"; exit 1 }
   - luarocks remove luv
+    # Test the alternate rockspec
+  - mkdir build\lib
+  - cp build\Release\uv.lib build\lib
+  - cp -a deps\libuv\include build
+  - ps: luarocks make rockspecs\$(ls rockspecs) LIBUV_DIR=build CFLAGS="/nologo /MT /O2"
+  - ps: if("$(Get-Location)" -eq $(lua -e "print(require'luv'.cwd())")) { "LuaRocks test OK" } else { "LuaRocks test failed"; exit 1 }
+  - luarocks remove luv
 
 artifacts:
   - path: luv.dll
--- a/bump.sh
+++ b/bump.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+version=$1
+if [ -z "$version" ]; then
+  echo "must specify a version" >&2
+  exit 1
+fi
+
+script="/^version/s@\"[^\"]\\+\"@\"${version}\"@"
+sed -e "${script}" -i luv-*.rockspec
+sed -e "${script}" -i rockspecs/luv-*.rockspec
+git mv luv-*.rockspec luv-${version}.rockspec
+git mv rockspecs/luv-*.rockspec rockspecs/luv-${version}.rockspec
+git add luv-${version}.rockspec rockspecs/luv-${version}.rockspec
--- a/luv-1.8.0-4.rockspec
+++ b/luv-1.8.0-4.rockspec
@@ -1,7 +1,7 @@
 package = "luv"
 version = "1.8.0-4"
 source = {
-  url = 'https://github.com/luvit/luv/releases/download/1.8.0-4/luv-1.8.0-4.tar.gz',
+  url = 'https://github.com/luvit/luv/releases/download/'..version..'/luv-'..version..'.tar.gz'
 }
 
 description = {
--- a/rockspecs/luv-1.8.0-4.rockspec
+++ b/rockspecs/luv-1.8.0-4.rockspec
@@ -1,0 +1,79 @@
+-- Alternate rockspec that uses luarocks builtin builder
+package = "luv"
+version = "1.8.0-4"
+source = {
+  url = 'https://github.com/luvit/luv/releases/download/'..version..'/luv-'..version..'.tar.gz'
+}
+
+description = {
+  summary = "Bare libuv bindings for lua",
+  detailed = [[
+libuv bindings for luajit and lua 5.1/5.2/5.3.
+
+This library makes libuv available to lua scripts. It was made for the luvit
+project but should usable from nearly any lua project.
+  ]],
+  homepage = "https://github.com/luvit/luv",
+  license = "Apache 2.0"
+}
+
+dependencies = {
+  "lua >= 5.1"
+}
+
+external_dependencies = {
+  LIBUV = {
+    header = 'uv.h'
+  }
+}
+
+local function make_modules()
+  return {
+    ['luv'] = {
+      sources = {'src/luv.c'},
+      libraries = {'uv'},
+      incdirs = {"$(LIBUV_INCDIR)"},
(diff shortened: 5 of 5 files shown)


creationix:
I actually like this, just a couple issues I foresee.

1. I'd love to support windows properly.  Currently luvi supports windows, linux, freebsd, and darwin.
2. I like it that a shared libuv can be used, but a statically linked libuv is my primary use case and needs to work well.  I don't think distribution libuv binaries are to the point they are usable much yet.  I tend to release a new luv the same day or day after a libuv release.

tarruda:
> I like it that a shared libuv can be used, but a statically linked libuv is my primary use case and needs to work well. I don't think distribution libuv binaries are to the point they are usable much yet. I tend to release a new luv the same day or day after a libuv release.

I don't think I can make luarocks build libuv automatically without invoking cmake and reintroducing the problems in #223, but I think the luarocks community is ok with installing their external dependencies manually.

Projects that use luv and would like to install libuv automatically can do so with a separate script, and by setting LIBUV_DIR appropriately(this is what neovim does).

tarruda:
@equalsraf I have copied the windows-specific bits from https://github.com/neovim/lua-client/pull/23, do you know how to fix [this error](https://ci.appveyor.com/project/racker-buildbot/luv/build/1.0.396/job/h9g08pm71o39m6gj#L1289)? (I think it fails in visual C 2013, but not 2015)

tarruda:
Fixed by replacing `/md` with `/mt` in windows CFLAGS, as explained in the CMakeLists.txt.

tarruda:
Fixed by replacing `/md` with `/mt` in windows CFLAGS, as explained in the CMakeLists.txt.

tarruda:
@creationix note that while the luarocks builtin builder doesn't allow to compile libuv automatically, the user can still build libuv using cmake and then invoke `luarocks make LIBUV_DIR=...`, as it was done in the travis/appveyor builds

creationix:
I understand that libuv can be built manually, but this is a regression for people wanting a static libuv inside luv.  Before they could simply install cmake and a compiler and do `luarocks install luv`, but now they have to also download and build libuv from source and install it somewhere and manually pass in the install path to luarocks.

I may be over-thinking the luarocks use case, but this is the area causing me pause and the reason in required cmake in the first place.

tarruda:
@creationix I understand the inconvenience of manually installing dependencies(which is why neovim build system automatically), but hiding cmake under luarocks has some disadvantages:

- It is not possible to specify the compiler to do the build
- It is not possible to pass arbitrary cmake flags

These limitations are introduced by luarocks not providing a way to customize the options passed to cmake beyond what is hardcoded in the rockspec. I may have missed something, but found no way to pass through arbitrary cmake options(such as `-DBUILD_SHARED_LIBS`) from the luarocks command line to cmake. Also asked a [question](https://github.com/keplerproject/luarocks/issues/509) but got no conclusive answer.

In the end it is a tradeoff between small gain in convenience for simple scenarios(like writing a script that uses luv) and making it more difficult/impossible to use in more complex scenarios(such as part of build system in a another project, that needs to customize cmake options beyond what luarocks allows). 

The reason I wrote this rockspec is because I couldn't find another way of doing what neovim build system needs for travis, but this PR not being merged does not block me because I can simply maintain this rockspec in neovim source tree, so it is up to your discretion. But before making a decision, ask yourself these questions:

- Does most luvit/luvi/luv community use luarocks? AFAIK luvit has its own package manager/module system, so I believe most users don't really care about luarocks
- Does the luarocks community care about automatically installed libraries?  Since luarocks provides no native way to automatically install libraries, I believe they are used and even prefer to take care of installing their own system dependencies.

tbastos:
I agree that not providing a way to customize the options passed to CMake is a serious luarocks limitation, but maybe we should fix luarocks instead? This PR's regression would make life harder for applications that use `luv` and are currently installed automatically using luarocks/cmake (admittedly I'm the author of one such application). On Linux the fix would be an easy `apt-get install`, but on Windows the amount of extra work to install my application would be immense. Fixing luarocks seems the nobler thing to do here, especially since big projects such as neovim already depend on CMake anyways.

tarruda:
> I agree that not providing a way to customize the options passed to CMake is a serious luarocks limitation, but maybe we should fix luarocks instead? This PR's regression would make life harder for applications that use luv and are currently installed automatically using luarocks/cmake (admittedly I'm the author of one such application). On Linux the fix would be an easy apt-get install, but on Windows the amount of extra work to install my application would be immense. Fixing luarocks seems the nobler thing to do here, especially since big projects such as neovim already depend on CMake anyways.

Agreed, but getting familiar with luarocks internals in order to submit a PR would take considerably longer, and unfortunately I don't have time to spare.

OTOH I already had a working rockspec from the previous version of neovim lua client, even the windows-specific were already implemented by @equalsraf, so it cost me only a few minutes to adapt the rockspec for luv.

creationix:
For now can we simply point people to @tarruda's rockspec if they want a more minimal spec?  I'm not familiar with rocks and how namespaces work.  Is it under a different name or a different repository?

tarruda:
> For now can we simply point people to @tarruda's rockspec if they want a more minimal spec? I'm not familiar with rocks and how namespaces work. Is it under a different name or a different repository?

To install luv with the "built" rockspec one can simply invoke `luarocks build https://raw.githubusercontent.com/tarruda/luv/18bc34aa72492fddc558641b86ca6fdad5fd7589/luv-1.8.0-4.rockspec LIBUV_DIR...`, but it won't install as a separate package(the rock will still be registered as "luv"). It is possible to change the name in the rockspec, but AFAIK other modules that depend on luv won't be able to use it without breaking dependency on luv.

How about this: I rename the file to luv-1.8.0-4-nodeps.rockspec and both are maintained in luv tree(only changing the version number will be necessary, but this can be automated with a bump script). Publishing to luarocks is not necessary, projects that require this minimal version can install by pointing to the github raw content.

Take neovim lua client for example, it depends on plain `luv`, and luarocks will install it using the default cmake rockspec. But if someone using neovim-client(eg: neovim) requires luv to be built using the "nodeps" rockspec, they can simply manually install luv prior to the client with: `luarocks build https://raw.githubusercontent.com/luv/luv/master/luv-1.8.0-4-nodeps.rockspec LIBUV_DIR...`

If this solution is acceptable I will fix this PR.

tarruda:
@creationix I have modified this PR: Now the modified rockspec is under the `rockspecs` directory and won't affect how luv is installed via luarocks. Projects that need to use the new rockspec can invoke luarocks passing the github raw url, eg: `luarocks build https://raw.githubusercontent.com/luv/luv/1.8.0-4/rockspecs/luv-1.8.0-4-nodeps.rockspec`.

The new rockspec also has to be updated when the version changes, but I have added a bump script to simplify the task.

creationix:
Awesome work!

PR #227, opened 2016-04-08
Clean up rockspec

daurnimator:
Using functions from a rockspec is silly.

Use a simple static table definition instead.
Things from the "platforms" section get merged in.


creationix:
looks good to me

creationix:
@tarruda do you have any objections with this change?

tarruda:
I was not aware that luarocks took care of recursively merging platform/common settings like that(In fact I simply copied from an existing rockspec). Is `nil` at the start of each platform-specific `libraries` table is a placeholder for `uv`?

In any case, I have no objections if nothing breaks, so LGTM

PR #235, opened 2016-06-08
Fix #204: sparse stdio in process:spawn()

### LINKED ISSUE #204: Process spawn() stdio can not specify UV_IGNORE with sparse table
The stdio option is handled as such:

```
options.stdio_count = len = lua_rawlen(L, -1);
```

...and the UV_IGNORE is specified as such:

```
else if (lua_type(L, -1) == LUA_TNIL) {
```

Due to how Lua handles sparse tables, setting STDIN (table element 1) to nil may trigger `lua_rawlen` to be set to 0 and thus making it impossible to specify STDOUT/STDERR when STDIN is set to nil. I've got this workaround that seems to work:

```
local function hack_stdio(stdio)
    local result
    if nil ~= stdio[3] then
        return {stdio[1], stdio[2], stdio[3]}
    end
    if nil ~= stdio[2] then
        return {stdio[1], stdio[2]}
    end
    if nil ~= stdio[1] then
        return {stdio[1]}
    end
    return {}
end
```

...but it would be much better if we could specify a "null" value that represents a file descriptor that we want to be set to UV_IGNORE. As a side-note, for interoperability, I've found it is best to check if a userdata has the __name metatable field set to "null" to determine if an object is null. I've found this gives the maximum interoperability.

See https://github.com/mpx/lua-cjson/issues/11#issuecomment-137562575

brimworks:
Define and use sparse_rawlen() for the stdio options passed into
process:spawn(). This fixes https://github.com/luvit/luv/issues/204

### DIFF
--- a/src/luv.h
+++ b/src/luv.h
@@ -59,6 +59,9 @@
 # define luaL_setfuncs(L,l,n) (assert(n==0), luaL_register(L,NULL,l))
 # define lua_resume(L,F,n) lua_resume(L,n)
 # define lua_pushglobaltable(L) lua_pushvalue(L, LUA_GLOBALSINDEX)
+# define lua_absindex(L, i)                              \
+    ((i) > 0 || (i) <= LUA_REGISTRYINDEX ?              \
+     (i) : lua_gettop(L) + (i) + 1)
 #endif
 
 /* There is a 1-1 relation between a lua_State and a uv_loop_t
--- a/src/process.c
+++ b/src/process.c
@@ -47,6 +47,29 @@ static void luv_clean_options(uv_process_options_t* options) {
   free(options->env);
 }
 
+// iterates over the tbl to find the max integer
+// key of the table that is >= 1. If any key is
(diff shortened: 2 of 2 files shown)


creationix:
Sorry I missed this notification.  This isn't the behavior I was going for, but I suppose it would be fine.  My only concern is if someone passed in a table like `{[1000000]=true}` that we'll allocate a giant buffer in C and then loop over a million nil FDs before finding one.

On the other hand, this is much simpler semantics than my proposal, so that's a plus.

If you still want this merged, please rebase it or if you're really busy but want it merged, I can rebase it for you.

brimworks:
I've rebased this against the master branch.

Regarding the concern for sparse tables, if someone really wanted to set file descriptor 1000000, then libuv requires allocating 1000000 uv_stdio_container_t structs.

I suppose we could impose a limit though. Normally 1024 is the limit, but it would be better if we could query the ulimits... just not sure how to do that in a platform agnostic way.

Thanks,
-Brian

PR #238, opened 2016-07-08
compile in ansi mode

yihuang:
Because I've compiled my lua as c++, I need to compile luv as c++ too. So I've fixed all the compilation problems with c++/ansi c compiler.
Just run ``gcc -ansi -Wall -c src/luv.c -Ideps/libuv/include`` , you can see current compile errors.


creationix:
Typecast all the things!

Issue #234, opened 2016-05-13
luv_check_callback NULL dereference


justinmk:
In https://github.com/neovim/neovim/pull/4743#issuecomment-218944511 we ran into a situation where `luv_check_callback` gets a null `data` argument. 

I notice that most of the `luv_check_*` functions (`luv_check_process`, `luv_check_poll`, `luv_check_idle`, `luv_check_signal`, ...) in luv check for `handle->data != NULL`. But `luv_check_stream` and `luv_check_handle` do not.

creationix:
Thanks, I'll see if I patch this.

creationix:
I added a quick sanity check, you'll probably get a different error now in whatever was triggering the segfault in neovim.

I'd still like to have a more robust type checking system, but this should at least prevent null deferences for the `luv_handle_t`.

justinmk:
Thanks!

PR #243, opened 2016-07-29
Make type checking for handles more robust. (#234)

creationix:
We need to make type checking more robust.  Currently closed handles don't count as handles and give unexpected type errors when calling methods like `handle:is_closing()` that are expected to be safe.


rphillips:
+1 lgtm

PR #247, opened 2016-11-10
fix compile errors as pure C code

uudiin:


uudiin:
Yes, Compiled by VS2008

creationix:
Thanks for the PR, sorry I let these in.  My windows workflow now has the newer compiler so I don't notice these issues anymore.

creationix:
At some point we might stop supporting vs2008, but as long as community members don't mind sending in PRs to fix the style I'll keep accepting them.

PR #250, opened 2016-12-16
fix other userdata casts

rphillips:
Cleans up the other places that do a dereference of the userdata


creationix:
+1
The issue with appveyor seems unrelated to this patch.
I was able to get travis passing again at least (upstream lua changes had broken everything)

PR #254, opened 2017-03-24
fix bug with wrong errfunc index, caused by https://github.com/luvit/…

zhaozg:
fix bug with wrong errfunc index.
Caused by my commit https://github.com/luvit/luv/commit/e5ec37cd6c16c963ce9158cf0b422a6276c030a9  very sorry for that


creationix:
Thanks for fixing this!

PR #256, opened 2017-04-13
examples: add pipe echo example

joerg-krause:
This is an luv implementation of the libuv book example for [pipes](https://nikhilm.github.io/uvbook/processes.html#pipes).


creationix:
Looks great, thanks!

Issue #257, opened 2017-05-04
cancel causes future SEGV


brimworks:
When I try to run `uv.cancel(req)`, when the callback associated with `req` is ran a segmentation fault occurs.

Per the documentation of uv_cancel():

http://docs.libuv.org/en/v1.x/request.html

...the request data should not be deallocated until the callback is ran, and all callbacks need to check for the canceled state.

My branch is pretty old, but looking at the latest code on master, it appears that request data deallocation is still happening when cancel is called:

https://github.com/luvit/luv/blob/master/src/req.c#L48

This is my fix, I can extract this into a pull request against master if desired:

https://github.com/brimworks/luv/commit/7425d8bea5278392fa1bf15bd68b2abe33726857

This seems to fix the SEGV for me and the request does actually get canceled.

Thanks,
-Brian

creationix:
That's awesome.  cancel was the one part of libuv I never figured out.  It wasn't a priority since nobody seemed to use it.

Yes, a PR would be welcome.

brimworks:
PR sent... as mentioned in the PR this is probably incomplete, but at least we don't get a SEGV :).

https://github.com/luvit/luv/pull/258

PR #258, opened 2017-05-04
Fix luv.cancel(req) so cleanup occurs in the callbacks

brimworks:
Fixes issue: 

https://github.com/luvit/luv/issues/257

NOTE: The documentation says cancel is implemented for a subset of request types:

http://docs.libuv.org/en/v1.x/request.html#c.uv_cancel

This change addresses some of the callbacks, but these probably also need to be changed to check for `UV_ECANCELED`:

- `luv_poll_cb()`
- `luv_connect_cb()`
- `luv_fs_event_cb()`
- `luv_fs_poll_cb()`
- `luv_shutdown_cb()`
- `luv_connection_cb()`
- `luv_write_cb()`
- `luv_udp_send_cb()`

We might want to also change the above functions so that the callbacks are not ran if `UV_ECANCELED` is the status. In light of the fact that the documentation claims cancel only works for a subset of types, not sure how critical it is avoid calling the callbacks when a request is canceled.

Thanks,
-Brian

### DIFF
--- a/src/req.c
+++ b/src/req.c
@@ -45,8 +45,7 @@ static int luv_cancel(lua_State* L) {
   uv_req_t* req = (uv_req_t*)luv_check_req(L, 1);
   int ret = uv_cancel(req);
   if (ret < 0) return luv_error(L, ret);
-  luv_cleanup_req(L, (luv_req_t*)req->data);
-  req->data = NULL;
+  // Cleanup occurs when callbacks are ran with UV_ECANCELED status.
   lua_pushinteger(L, ret);
   return 1;
 }


creationix:
Hmm, looking at this closer, I think the desired behavior is to call the callback in lua if libuv calls the callback in C and make sure the error value contains the canceled message.

So now I'm wondering why this didn't just work already.  I already check for a non zero status and report the failure.

creationix:
I wonder if `UV_ECANCELED` is positive and I'm checking for for `(status < 0)` and that's why the code is not catching it?

If that's the case *and* we want to call the lua callback, then simply changing `(status < 0)` to `(status)` for those functions would fix it right?

Now some functions have meaning in positive status so we can't blanket change them all.

brimworks:
The most important change is to req.c to avoid the SEGV... otherwise the data is deallocated and set to null... then in the callback the data is dereferenced.

brimworks:
IMHO, if I cancel a task I don't expect the callback to be ran... but this is simply a question of what semantics you want for `uv.cancel()`.

creationix:
I want cancel called, because often I use luv with coroutines.

My coroutine is suspended when the call starts and it's resumed when the callback comes back.  If you never call the callback the code will hang forever.

Also it's common in callback code to assume the callback will always be called.

brimworks:
You're right. Better to still run the callback. Currently I only have access to a phone, so I can't update the pull request until tomorrow.

creationix:
I'm looking to do a luv release later today, if you won't have time to update this, I can see about finishing it myself.

brimworks:
I've updated the pull request so it simply removes the deallocation code.

Thanks,
-Brian

PR #259, opened 2017-05-05
Update all the things

creationix:
Updating dependencies.

I still need to go through the libuv changelog and see if there are new interfaces that need binding.


creationix:
Looks like v1.10.0 added a function.

https://abi-laboratory.pro/tracker/compat_report/libuv/1.9.1/1.10.0/121f9/abi_compat_report.html#Added

```c
uv_translate_sys_error ( int sys_errno )
```

creationix:
I'm not sure the new API needs binding in lua, leaving it out for now.

Now if we can just get the CI green again.

zhaozg:
hi @creationix please look at https://github.com/luvit/luv/pull/262

PR #262, opened 2017-06-23
update luv to support latest libuv

zhaozg:


zhaozg:
please check at https://github.com/libuv/libuv/blob/v1.x/docs/src/misc.rst#api and https://github.com/libuv/libuv/blob/v1.x/docs/src/errors.rst#api

creationix:
oh nice, they added environment variable functions!

jdonaldson:
It would be nice to get a version bump for luv.  I can drop a separate environment management library from my project dependencies.

creationix:
Good point, I had forgot about this.  I'm heading to Houston to help with volunteer flood cleanup this weekend so I'm not sure I'll get a build out.  If I don't, feel free to ping me again next week.

jdonaldson:
Oh wow, don't worry about this then.  Good luck over there, I have friends and family down there too.  Thanks for helping out!

PR #266, opened 2017-10-31
Allow callable tables as callback parameters (closes #265)

### LINKED ISSUE #265: __call support
Does luv support __call cbs?

squeek502:
`pcall` (which is used to handle the actual calling of the callbacks) handles `__call` already, so the only thing blocking this was the `LUA_TFUNCTION` type check.

Example code that will now work:
```lua
local callable = setmetatable({}, {__call=function(...) print("called", ...) end})

callable()
pcall(callable)

local uv = require('uv')

local handle = uv.new_timer()
local delay = 0
uv.timer_start(handle, delay, 0, callable)

uv.run()
```
Which outputs:
```
called  table: 0x000d8420
called  table: 0x000d8420
called  table: 0x000d8420
```

Example error when trying to pass something non-callable as a callback argument:
```
bad argument #4 to 'timer_start' (function or callable table expected, got boolean)
```

(related: #265)

### DIFF
--- a/src/lhandle.c
+++ b/src/lhandle.c
@@ -53,8 +53,34 @@ static luv_handle_t* luv_setup_handle(lua_State* L) {
   return data;
 }
 
+static int luv_is_callable(lua_State* L, int index) {
+  if (luaL_getmetafield(L, index, "__call") != LUA_TNIL) {
+    // getmetatable(x).__call must be a function for x() to work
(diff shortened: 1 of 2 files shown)


creationix:
This is exactly what I hoped for.  I just didn't remember if I had implemented it yet.

Thanks!

PR #270, opened 2017-12-02
recv callback expects 4 arguments

samunders-core:


src/udp.c
```diff
@@ -210,7 +210,7 @@ static void luv_udp_recv_cb(uv_udp_t* handle, ssize_t nread, const uv_buf_t* buf
 
   // err
   if (nread < 0) {
-    luv_status(L, nread);
+    luv_status(L, nread); // TODO: I think luv_push_status_nil(L, nread) would be better name => remove branching
```
rphillips: we need to remove this comment from the PR

PR #283, opened 2018-04-01
Link the libraries libpthread and libc++abi

utkarsh009:
These are required for building LuaJIT on OpenBSD with clang. They provide the symbols `_Unwind_*`.

### DIFF
--- a/deps/uv.cmake
+++ b/deps/uv.cmake
@@ -114,7 +114,6 @@ else()
     ${LIBUVDIR}/src/unix/pipe.c
     ${LIBUVDIR}/src/unix/poll.c
     ${LIBUVDIR}/src/unix/process.c
-    ${LIBUVDIR}/src/unix/proctitle.c
     ${LIBUVDIR}/src/unix/signal.c
     ${LIBUVDIR}/src/unix/spinlock.h
     ${LIBUVDIR}/src/unix/stream.c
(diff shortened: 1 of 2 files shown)


utkarsh009:
According to [this](https://github.com/libuv/libuv/blob/9f07a3673b301da99bd31f2d793983caad6e3428/Makefile.am#L430), `openbsd.c` already has the function definitions that `proctitle.c` provides. Same is the case with `freebsd.c` and `netbsd.c`. The linked file doesn't include `proctitle.c` in any of the *BSDs. Although, Darwin includes it.

deps/uv.cmake
```diff
@@ -114,7 +114,6 @@ else()
     ${LIBUVDIR}/src/unix/pipe.c
     ${LIBUVDIR}/src/unix/poll.c
     ${LIBUVDIR}/src/unix/process.c
-    ${LIBUVDIR}/src/unix/proctitle.c
```
zhaozg: This should keep.
156th line(proctitle.c) in linux section should remove.

deps/uv.cmake
```diff
@@ -181,6 +180,7 @@ if(APPLE)
     -D=_DARWIN_USE_64_BIT_INODE
   )
   set(SOURCES ${SOURCES}
+    ${LIBUVDIR}/src/unix/proctitle.c
```
zhaozg: This should remove.

utkarsh009:
I don’t have OpenBSD VM right now to test it. But, removing it from the block where CMAKE_SYSTEM_NAME is Linux, won’t make any difference in OpenBSD. proctitle.c is required only in a few OSes. By adding that file to SOURCES, for every non WIN32 OS, you are also adding it to OpenBSD, so I had to remove it from there. But, MacOS required it, so I had to explicitly add it there. If any of the other UNIX like OSes require it, you will have to add it separately in that block.

zhaozg:
thank you for the explanation

utkarsh009:
You’re welcome. 😇

PR #286, opened 2018-04-24
Fixup about submodule lua-compat-5.3, against to 34ada3e1d75796d2295e…

zhaozg:
Fixup about submodule lua-compat-5.3, against to 34ada3e1d75796d2295e…

And this will better support luvi


zhaozg:
ci failed here https://travis-ci.org/luvit/luv/jobs/370449135#L375, but not related the PR.

PR #289, opened 2018-05-21
Use system sleep() function

mnemnion:
The local `sleep` var in `hare_id` was assigned but never used. 

Tested this, works fine on macOS, although the field value is milliseconds, not seconds.


mnemnion:
The example brings in the system sleep function, but then doesn't use it.

The tortoise thread uses uv.sleep, without creating a local sleep.

Imho the tutorial would read fine with one sleep instead of three, although
I like that it illustrates using the FFI inside the thread.

But invoking three and using only one is confusing.

zhaozg:
thanks

PR #291, opened 2018-07-09
build: smarter LUA_add_custom_commands

zhaozg:


zhaozg:
for https://github.com/luvit/luvi/issues/167

squeek502:
This fails on Windows, see https://ci.appveyor.com/project/racker-buildbot/luvi/branch/master/job/lbnl00gj7bvjexqy

`if(NOT file MATCHES "^/")` will fail to detect absolute paths on Windows (which typically start with a drive letter like `C:/`, but can also have a network path like `\\server\share`).

Instead, [`get_filename_component with REALPATH`](https://cmake.org/cmake/help/v3.4/command/get_filename_component.html) or [`IS_ABSOLUTE`](https://cmake.org/cmake/help/v3.4/command/if.html?highlight=is_absolute) should be used.

Issue #275, opened 2018-02-08
Please update the Luarocks package


slact:
The package on Luarocks.org hasn't been updated in over a year and is missing some critical fixes. Please release an up-to-date version so that my own luarocks don't have to include and build luv "by hand".

justinmk:
@creationix Any plans for a release?

creationix:
I'd love to.  I've just been crazy busy.  Ping me next week and I'll see if I can get a new release of everything out.

cliffordjefferies:
Having an issue building - 
sudo luarocks install luv
Installing https://rocks.moonscript.org/luv-1.22.0-0.src.rock...
Using https://rocks.moonscript.org/luv-1.22.0-0.src.rock... switching to 'build' mode
-- The C compiler identification is GNU 5.4.0
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- The CXX compiler identification is GNU 5.4.0
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Lua: using information from luarocks
-- LUA_LIBDIR: /usr/local/lib
-- LUA_INCDIR: /usr/include/lua5.1
-- LUA: /usr/bin/lua5.1
-- Lua library: LUA_LIBRARIES-NOTFOUND
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/luarocks_luv-1.22.0-0-1939/luv-1.22.0-0
Scanning dependencies of target uv
[  1%] Building C object deps/libuv/CMakeFiles/uv.dir/src/fs-poll.c.o
[  3%] Building C object deps/libuv/CMakeFiles/uv.dir/src/inet.c.o
[  4%] Building C object deps/libuv/CMakeFiles/uv.dir/src/threadpool.c.o
[  6%] Building C object deps/libuv/CMakeFiles/uv.dir/src/timer.c.o
[  7%] Building C object deps/libuv/CMakeFiles/uv.dir/src/uv-common.c.o
[...]

PR #300, opened 2018-08-17
Include deps/lua-compat-5.3 in luarocks .tar.gz

squeek502:
Should fix Luarocks build error reported here: https://github.com/luvit/luv/issues/275#issuecomment-413751038


zhaozg:
+1

Issue #301, opened 2018-08-25
update libuv to 1.23.0


zhaozg:
Add new API for luv

* fs: add uv_open_osfhandle (Bartosz Sosnowski)
* win: add uv__getnameinfo_work() error handling (A. Hauptmann)
* unix,win: add uv_os_{get,set}priority() (cjihrig)

PR #302, opened 2018-08-25
libuv: bump to 1.23.0

rphillips:
We do not export uv_get_osfhandle or uv_open_osfhandle yet...

Fixes: https://github.com/luvit/luv/issues/301


zhaozg:
* uv__getnameinfo_work are low level IMP for uv_getnameinfo, which already in https://github.com/libuv/libuv/blob/90891b4232e91dbd7a2e2077e4d23d16a374b41d/src/win/getnameinfo.c#L117
* uv_get_osfhandle  and uv_open_osfhandle are os relative C api, and the are not need in lua.

So +1 for merge

PR #308, opened 2018-10-30
Fix checking if environment variable is defined

joerg-krause:
The correct way to check if an environment variable is defined is to use `if(DEFINED ENV{})`.

See: https://cmake.org/pipermail/cmake/2011-October/046704.html


joerg-krause:
Any feedback on this?

PR #316, opened 2019-02-20
Update libuv to 1.26.0

squeek502:
- uv.new_thread can now take a table as an optional first argument with the optional key "stack_size" that specifies the stack size for the new thread (e.g. `uv.new_thread({stack_size=5}, function() end)`); see https://github.com/libuv/libuv/commit/0eca049a9b3ca7440b4a4c00ab0ebe57cc3ff948
- Bumped luv version to 1.26.0 to match


zhaozg:
+1

PR #318, opened 2019-03-19
Update to Libuv 1.27.0

squeek502:
Additions:
 - uv.udp_getpeername(handle)
 - uv.udp_connect(handle, host, port) [called with nil host and nil port to clear the connection]

Changes:
 - uv.udp_send() and uv.udp_try_send() to accept connected udp handles (via udp_connect); if a connected handle is used, then the host and port args must be nil

---

For `udp_send`/`udp_try_send`: host and port are separate arguments in the Luv bindings, so I made using a connected handle require both arguments to be nil. However, it would be possible to skip the port argument if the host argument is nil (so `udp_send(handle, data, nil, callback)` vs `udp_send(handle, data, nil, nil, callback`). Not sure which one would be better.

---

Libuv tag comparison: https://github.com/libuv/libuv/compare/v1.26.0...v1.27.0
Libuv changelog: https://github.com/libuv/libuv/blob/23bb992d7b38a83bb83fb0851148d9f6941157f0/ChangeLog


src/luv.c
```diff
@@ -436,6 +440,10 @@ static const luaL_Reg luv_udp_methods[] = {
   {"try_send", luv_udp_try_send},
   {"recv_start", luv_udp_recv_start},
   {"recv_stop", luv_udp_recv_stop},
+#if LUV_UV_VERSION_GEQ(1, 27, 0)
+  {"connect", luv_udp_connect},
```
rphillips: Shouldn't this be udp_connect and udp_getpeername on the next line?

squeek502: Those are the metatable bindings for uv_udp_t. So it's for `handle:connect()`, `handle:getpeername()`. The prefixed versions are here: https://github.com/luvit/luv/pull/318/files#diff-8792371db05ee9c0a8a1edb4d4f421acR183

PR #327, opened 2019-05-30
Refactoring the code to get the loop

zhaozg:
use luv_req_t and luv_handle_t struct filed to hold lua_State pointer,
uv_loop_t data filed may be used by outside of luv

Ref discuss of https://github.com/luvit/luv/pull/325

### DIFF
--- a/src/async.c
+++ b/src/async.c
@@ -24,8 +24,8 @@ static uv_async_t* luv_check_async(lua_State* L, int index) {
 }
 
 static void luv_async_cb(uv_async_t* handle) {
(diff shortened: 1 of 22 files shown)


andreypopp:
LGTM.

Does this mean it's a breaking change? (remove of a function from `luv.h`).

zhaozg:
> Does this mean it's a breaking change? (remove of a function from `luv.h`).

Yes, C API break(not the focus), but not lua API (key point)

zhaozg:
If other luvit members not active,  I'll release a new 1.29.1-1 version in next 12 hours.

PR #331, opened 2019-06-01
Implement support for external uv_loop_t

### LINKED ISSUE #244: Segmentation fault when executing busted with to test files bot importing 'luv'
See this: https://gist.github.com/AitorATuin/46bafaeeb51388c6c1d367083b83c4c6

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff6637a92 in uv_walk () from /home/eof/.luarocks/lib/lua/5.1/luv.so
(gdb) bt
#0  0x00007ffff6637a92 in uv_walk () from /home/eof/.luarocks/lib/lua/5.1/luv.so
#1  0x00007ffff663394a in loop_gc () from /home/eof/.luarocks/lib/lua/5.1/luv.so
#2  0x0000000000408058 in ?? ()
#3  0x0000000000408454 in ?? ()
#4  0x000000000040942b in ?? ()
#5  0x000000000040a2b8 in ?? ()
#6  0x00000000004077bb in ?? ()
#7  0x000000000040ddd2 in lua_close ()
#8  0x0000000000403cda in main ()
```

zhaozg:
* Implement support for external uv_loop_t, many thanks to @andreypopp
* fix #244

### DIFF
--- a/src/luv.c
+++ b/src/luv.c
@@ -567,6 +567,8 @@ static void walk_cb(uv_handle_t *handle, void *arg)
 
 static int loop_gc(lua_State *L) {
   uv_loop_t* loop = luv_loop(L);
+  if (loop==NULL)
+    return 0;
   // Call uv_close on every active handle
   uv_walk(loop, walk_cb, NULL);
   // Run the event loop until all handles are successfully closed
--- a/src/luv.c
+++ b/src/luv.c
@@ -508,18 +508,53 @@ static void luv_handle_init(lua_State* L) {
   lua_setfield(L, LUA_REGISTRYINDEX, "uv_stream");
 }
 
-LUALIB_API lua_State* luv_state(uv_loop_t* loop) {
-  return (lua_State*)loop->data;
+// TODO: see if we can avoid using a string key for this to increase performance
+static const char* luv_loop_key = "luv_loop";
+static const char* luv_state_key = "luv_main_thread";
+
+// Get main thread, ensure coroutines works
+// Only called when luv setup uv_handle_t or uv_req_t
+LUALIB_API lua_State* luv_state(lua_State* L) {
+  lua_State* lstate;
+  lua_pushstring(L, luv_state_key);
+  lua_rawget(L, LUA_REGISTRYINDEX);
+  lstate = lua_tothread(L, -1);
+  lua_pop(L, 1);
+  if (lstate==NULL) {
+    luaL_error(L, "cannot get main thread");
+  }
+  return lstate;
 }
 
-// TODO: find out if storing this somehow in an upvalue is faster
+// Called when luv init uv_handle_t
 LUALIB_API uv_loop_t* luv_loop(lua_State* L) {
   uv_loop_t* loop;
-  lua_pushstring(L, "uv_loop");
+  lua_pushstring(L, luv_loop_key);
   lua_rawget(L, LUA_REGISTRYINDEX);
-  loop = (uv_loop_t*)lua_touserdata(L, -1);
+  if (lua_isnil(L, -1))
+    loop = NULL;
+  else
+    loop = *(uv_loop_t**)lua_touserdata(L, -1);
   lua_pop(L, 1);
   return loop;
+}
+
+// Set an extran loop, before luaopen_luv
+LUALIB_API void luv_set_loop(lua_State* L, uv_loop_t* loop) {
+  if (loop==NULL) {
+    lua_pushstring(L, luv_loop_key);
+    lua_pushnil(L);
+    lua_rawset(L, LUA_REGISTRYINDEX);
+  } else {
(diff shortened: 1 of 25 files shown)


zhaozg:
@squeek502 can you leave you options about this PR, this is a quite complicated change.

zhaozg:
split it to some PRs https://github.com/luvit/luv/pull/335

src/luv.c
```diff
@@ -506,20 +512,50 @@ static void luv_handle_init(lua_State* L) {
...
+    luaL_error(L, "cannot get main thread");
+  }
+  return lstate;
 }
 
-// TODO: find out if storing this somehow in an upvalue is faster
 LUALIB_API uv_loop_t* luv_loop(lua_State* L) {
   uv_loop_t* loop;
-  lua_pushstring(L, "uv_loop");
+  lua_pushlightuserdata(L, (void*)luv_loop_key);
```
squeek502: Maybe keep this as `pushstring` to avoid light userdata? Same with `luv_state_key`.

zhaozg: Ok. Is there a more quick way? those api are repeat loop a large numbers times.

squeek502: Is there a reason it needs to be retrieved from the Lua registry each time? Could we store a pointer to the loop in C as well and use that?

zhaozg:
A brief description

* lua_State attached in loop->data changed to uv_handle_t->data->L or uv_req_t->data->, loop->data maybe used by other libs that manage extern loop, so we store L(lua vm) in luv private scope.
* In luv_setup_req and luv_setup_handle, we use luv_state() to get L(main thread), which is L call luv_set_loop or luaopen_luv.  L(main thread) in registry table with luv_state_key key.
* uv_loop_t in registry table with luv_loop_key key, both support luv own loop or extern loop.

src/luv.c
zhaozg: Not each time,  mainly when luv_setup_req or luv_setup_handle,

src/luv.c
```diff
@@ -508,20 +508,54 @@ static void luv_handle_init(lua_State* L) {
   lua_setfield(L, LUA_REGISTRYINDEX, "uv_stream");
 }
 
-LUALIB_API lua_State* luv_state(uv_loop_t* loop) {
-  return (lua_State*)loop->data;
+static const char* luv_loop_key = "uv_loop";
+static const char* luv_state_key = "main_thread";
```
squeek502: This should be changed to `"luv_main_thread"` to make sure it never collides with anything else.

src/luv.c
```diff
@@ -508,20 +508,54 @@ static void luv_handle_init(lua_State* L) {
...
-  loop = (uv_loop_t*)lua_touserdata(L, -1);
+  if (lua_isnil(L, -1))
+    loop = NULL;
+  else
+    loop = *(uv_loop_t**)lua_touserdata(L, -1);
   lua_pop(L, 1);
   return loop;
 }
 
+// Set an extran loop, before luaopen_luv
```
squeek502: Typo: this should say "external loop"

zhaozg:
> Using `luaL_ref` to store the lua_State in the registry might speed things up but I think we'd have to store an internal map of `lua_State* -> int ref` pairings to do that.

How about use [LUA_RIDX_MAINTHREAD](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_RIDX_MAINTHREAD), that first appeared in lua-5.3, and will make it compat to lua 5.3

squeek502:
> How about use [LUA_RIDX_MAINTHREAD](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_RIDX_MAINTHREAD), that first appeared in lua-5.2, and will make it compat to lua 5.2/5.2

I don't think we can do this. It will overwrite a ref value if it already exists at index 1, and it will if any other modules call `luaL_ref`.

In Lua 5.2/5.3, `LUA_RIDX_MAINTHREAD`'s index is set in `init_registry` to make sure that all future ref values are larger (the first call to `luaL_ref` will get `3`). In Lua 5.1, the first call to `luaL_ref` will get `1`.

squeek502:
By the way, I don't think the performance impact is a big issue. I think using `pushstring` is okay for now, maybe just add a comment like `// TODO: see if we can avoid using a string key for this to increase performance`.

zhaozg:
> By the way, I don't think the performance impact is a big issue. I think using `pushstring` is okay for now, maybe just add a comment like `// TODO: see if we can avoid using a string key for this to increase performance`.

YES, I think we shoud make performance tuning in others PR.

zhaozg:
> `// TODO: see if we can avoid using a string key for this to increase performance`.

here: https://github.com/luvit/luv/pull/331/commits/8a88b329ae6d7e9dd925429510a4dece9421900f#diff-8792371db05ee9c0a8a1edb4d4f421acR511

PR #335, opened 2019-06-03
Fix steps

### LINKED ISSUE #244: Segmentation fault when executing busted with to test files bot importing 'luv'

zhaozg:
* fix warning about unused static function
* fix ARM64 (aarch64) Lua "bad light userdata pointer"  …

### DIFF
--- a/src/luv.c
+++ b/src/luv.c
@@ -18,6 +18,7 @@
 #if (LUA_VERSION_NUM != 503)
 #include "c-api/compat-5.3.h"
 #endif
+#define LUV_SOURCE
 #include "luv.h"
 #include "util.c"
 #include "lhandle.c"
--- a/src/luv.c
+++ b/src/luv.c
@@ -20,6 +20,7 @@
 #endif
 #define LUV_SOURCE
 #include "luv.h"
+
 #include "util.c"
 #include "lhandle.c"
 #include "lreq.c"
--- a/src/async.c
+++ b/src/async.c
@@ -26,7 +26,7 @@ static uv_async_t* luv_check_async(lua_State* L, int index) {
 static void luv_async_cb(uv_async_t* handle) {
   lua_State* L = luv_state(handle->loop);
   luv_handle_t* data = (luv_handle_t*)handle->data;
-  int n = luv_thread_arg_push(L, (const luv_thread_arg_t*)data->extra, 0);
+  int n = luv_thread_arg_push(L, (luv_thread_arg_t*)data->extra, 0);
   luv_call_callback(L, data, LUV_ASYNC, n);
   luv_thread_arg_clear(L, (luv_thread_arg_t*)data->extra, 0);
 }
--- a/src/lhandle.h
+++ b/src/lhandle.h
@@ -45,6 +45,7 @@ typedef struct {
   void* extra;
 } luv_handle_t;
 
+#ifdef LUV_SOURCE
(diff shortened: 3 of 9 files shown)


zhaozg:
Split from https://github.com/luvit/luv/pull/331, to reduce complexity

src/luv.c
```diff
@@ -18,7 +18,13 @@
 #if (LUA_VERSION_NUM != 503)
 #include "c-api/compat-5.3.h"
 #endif
+#define LUV_SOURCE
 #include "luv.h"
+
+#if defined(__aarch64__)
+#define lua_pushlightuserdata(L, P)  lua_pushinteger((L), (lua_Integer)(P))
```
squeek502: Is this a complete enough fix? `lua_islightuserdata` will not work, for example. Maybe we should just avoid using light userdata instead?

The only places its used right now are:
- https://github.com/luvit/luv/blob/9e6e4f2f98647005d5dc0154ca9161540e73f499/src/thread.c#L118-L129
- https://github.com/luvit/luv/blob/9e6e4f2f98647005d5dc0154ca9161540e73f499/src/work.c#L174-L177
- https://github.com/luvit/luv/blob/9e6e4f2f98647005d5dc0154ca9161540e73f499/src/work.c#L260-L263

zhaozg: I think that's enough, because  we use use lua_pushlightuserdata just as a key for registry table, and not use it as a real pointer or userdata,  lua_pushlightuserdata is keep in luv private source code. As for lua_islightuserdata, we never use it.
The best solution is to remove lightuserdata api from luv, but that large change,  so I use define to get same result.

zhaozg: wait here https://github.com/luvit/luv/blob/9e6e4f2f98647005d5dc0154ca9161540e73f499/src/thread.c#L72-L74 maybe wrong.

zhaozg:
I have removed all lightuserdate from luv.  @squeek502

squeek502:
Is there a reason we aren't using `luaL_ref`/`luaL_unref` for this?

From https://www.lua.org/pil/27.3.2.html:

> You should never use numbers as keys in the registry, because such keys are reserved for the reference system. This system is composed by a couple of functions in the auxiliary library that allow you to store values in the registry without worrying about how to create unique names. (Actually, those functions can act on any table, but they are typically used with the registry.)

squeek502:
@rphillips how would a major version bump work for the versioning luv uses (i.e. right now it basically just mirrors the version of the libuv it binds)?

PR #345, opened 2019-06-24
allow embedder to control the execution and error handling of callbacks

bfredl:
For effecient use in neovim we would need a bit of control how luv executes callbacks. 

Primarily we need to take control of handling of errors. Currently luv prints errors in callbacks directly to stderr which messes up the internal screen of nvim. But we would also like to set a flag around the execution of the callback, as some lua API functions in nvim need to behave differently inside a luv callback (to avoid nested event loop for instance).

My suggestion is to add a "hook" function that a C embedder can set in a similar way to setting the event loop. Then the embedder is responsible for executing the callback and dealing with errors, including potential formatting of tracebacks etc. https://github.com/neovim/neovim/pull/10316/commits/aa78c74b74c6a0e3595b76d02f674dacabd6e092 is an example of usage.

Does this seem reasonable? if so I can clean up the patch, and add docs etc.

### DIFF
--- a/src/lhandle.c
+++ b/src/lhandle.c
@@ -116,15 +116,20 @@ static void luv_call_callback(lua_State* L, luv_handle_t* data, luv_callback_id 
   }
   else {
     int errfunc, ret;
+    luv_cb_hook_t hook = luv_cb_hook(L);
 
-    // Get the traceback function in case of error
-    lua_pushcfunction(L, traceback);
-    errfunc = lua_gettop(L);
-    // And insert it before the args if there are any.
-    if (nargs) {
-      lua_insert(L, -1 - nargs);
-      errfunc -= nargs;
+    if (hook == NULL) {
+      // Get the traceback function in case of error
+      lua_pushcfunction(L, traceback);
+      errfunc = lua_gettop(L);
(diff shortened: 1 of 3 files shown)


bfredl:
I guess a concern could be too many small userdata like both `luv_loop_key` and `luv_cb_hook_key`. A refactor could be to instead have a single userdata with a struct of the "global" luv state of a lua state.

zhaozg:
> Primarily we need to take control of handling of errors. Currently luv prints errors in callbacks directly to stderr which messes up the internal screen of nvim. 

How a about generate a error with lua_error

> My suggestion is to add a "hook" function that a C embedder can set in a similar way to setting the event loop. 

I have no good idea,  want to hear the voices of others.

bfredl:
> How a about generate a error with lua_error

How would this work? `lua_error` only works inside a cfunction called from lua, and here we are outside any lua stack.

zhaozg:
Do you mind if I re-implement this feature? That will break the API you defined. @bfredl

bfredl:
The API was only a suggestion, Yours also handles return values, so it is better.

PR #347, opened 2019-06-26
Allow dns tests that require an internet connection to fail when offline

### LINKED ISSUE #340: tests/run.lua ... allow exclusion of network-requiring tests
When building the package for Linux distributions, it is usually run in the environment which doesn't have network access. Would it be possible to update test running mechanism to allow exclusion of all tests marked as requiring network access. I now remove file ``tests/test_dns.lua``, but that's certainly not a good solution.

squeek502:
Closes #340

This allows the full `make test` suite to run while offline.

### DIFF
--- a/tests/test-dns.lua
+++ b/tests/test-dns.lua
@@ -19,6 +19,11 @@ return require('lib/tap')(function (test)
       socktype = "stream",
       family = "inet",
     }, expect(function (err, res)
+      -- allow failure when offline
+      if err == "EAI_AGAIN" then
+        print(err, "skipping")
+        return
+      end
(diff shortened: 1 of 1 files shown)


zhaozg:
How about add an env `LUV-SKIP-INTERNET-TEST` to avoid test access internet

squeek502:
I think allowing failure here is okay:
- It means that `make test` will 'just work' even while offline
- The tests are repeated for local addresses, so we shouldn't be losing much coverage
- The allowed error is somewhat specific to having no network connection, it (hopefully) shouldn't fail with `EAI_AGAIN` when online

PR #350, opened 2019-06-27
support custom/external event callback

zhaozg:
An alternative way to achieve https://github.com/luvit/luv/pull/345

### DIFF
--- a/src/luv.c
+++ b/src/luv.c
@@ -511,39 +511,53 @@ static void luv_handle_init(lua_State* L) {
 // Call lua function, will pop nargs values from top of vm stack and push some
 // values according to nresults. When error occurs, it will print error message
 // to stderr, and memory allocation error will cause exit.
-static int luv_cfpcall(lua_State* L, int nargs, int nresult) {
-  int ret, errfunc;
+LUALIB_API int luv_cfpcall(lua_State* L, int nargs, int nresult, int flags) {
+  int ret, top, errfunc;
 
   // Get the traceback function in case of error
-  lua_pushcfunction(L, luv_traceback);
-  errfunc = lua_gettop(L);
-  // And insert it before the function and args
-  lua_insert(L, -2 - nargs);
-  errfunc -= (nargs+1);
+  if ((flags & (LUVF_CALLBACK_NOTRACEBACK|LUVF_CALLBACK_NOERRMSG) ) == 0)
+  {
+    lua_pushcfunction(L, luv_traceback);
+    errfunc = lua_gettop(L);
+    // And insert it before the function and args
+    lua_insert(L, -2 - nargs);
+    errfunc -= (nargs+1);
+  }else
+    errfunc = 0;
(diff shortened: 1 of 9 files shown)


src/luv.c
```diff
@@ -508,6 +508,39 @@ static void luv_handle_init(lua_State* L) {
...
+  case LUA_ERRMEM:
+    fprintf(stderr, "System Error: %s\n", lua_tostring(L, -1));
+    exit(-1);
+    break;
+  case LUA_ERRRUN:
+  case LUA_ERRSYNTAX:
+  case LUA_ERRERR:
+  default:
+    fprintf(stderr, "Uncaught Error: %s\n", lua_tostring(L, -1));
+    lua_pop(L, 1);
```
bfredl: shouldn't we return `0` in case of error, as no results will be pushed on the stack? or maybe `-1` as an indicator.  (doesn't matter now, but in case this is used with return values later, it is good to make the interface proper).

zhaozg:
ready to merge

src/luv.c
```diff
@@ -508,6 +508,56 @@ static void luv_handle_init(lua_State* L) {
...
+  }else
+    errfunc = 0;
+  top  = lua_gettop(L);
+
+  ret = lua_pcall(L, nargs, nresult, errfunc);
+  switch (ret) {
+  case LUA_OK:
+    break;
+  case LUA_ERRMEM:
+    fprintf(stderr, "System Error: %s\n", lua_tostring(L, -1));
```
rphillips: these fprintf's on 534 and 544 should be wrapped as well.

zhaozg: > should be wrapped
Sorry don't know your means, please give me an example.

src/luv.c
```diff
@@ -508,6 +508,56 @@ static void luv_handle_init(lua_State* L) {
...
+    if ((flags & LUVF_CALLBACK_NOEXIT) == 0)
+      exit(-1);
+    lua_pop(L, 1);
+    ret = -ret;
+    break;
+  case LUA_ERRRUN:
+  case LUA_ERRSYNTAX:
+  case LUA_ERRERR:
+  default:
+    fprintf(stderr, "Uncaught Error: %s\n", lua_tostring(L, -1));
```
rphillips: We could create another flag or reuse the `LUVF_CALLBACK_NOTRACEBACK` flag:

```
if ((flags & LUVF_CALLBACK_NOTRACEBACK) == 0) {
 fprintf(stderr, "Uncaught Error: %s\n", lua_tostring(L, -1));
}
```

zhaozg: OK I think need a new flag LUVF_CALLBACK_NOERRMSG

zhaozg: LUVF_CALLBACK_NOERRMSG will make callback don't traceback and not output any error message, but LUVF_CALLBACK_NOTRACEBACK only don't traceback to obtain more error message, but output origin message to stderr.

PR #352, opened 2019-06-29
update libuv to v1.30.0

zhaozg:


zhaozg:
After merge it, we can release v1.30.0-0

PR #359, opened 2019-07-10
build: cmake: shared: use REQUIRED with find_package(Libuv)

blueyed:
This is meant to improve the error message, which previously would look
like this:

    -- Could NOT find LIBUV (missing:  LIBUV_LIBRARIES)
    -- Found LUAJIT: /home/travis/nvim-deps/usr/lib/libluajit-5.1.so
    CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
    Please set them or make sure they are set and tested correctly in the CMake files:
    LIBUV_LIBRARIES
        linked by target "luv" in directory /home/travis/nvim-deps/build/src/luv


blueyed:
`LIBUV_LIBRARIES` is used always: https://github.com/blueyed/luv/blob/0129fe96e2d8ca5530dc2ffe2665f9dc5f935ee5/CMakeLists.txt#L160-L179

PR #380, opened 2019-09-21
Fix loop __gc not being called

### LINKED ISSUE #379: Memory leak (loop was never GCed)
Compiled luv.so with CLANG's address sanitizer, and run the unit tests, and found lots of leaks.

Troubleshooted a bit and found that the luv.c:loop_gc() was never triggered.

The issue seems started from revision db55f6bd.

    lua_pushstring(L, "_loop");
    loop = (uv_loop_t*)lua_newuserdata(L, sizeof(*loop));
    lua_rawset(L, -3);  // ref to loop, avoid __gc early

If remove the pushstring and rawset, loop_gc would be invoked too early, just as the comment says. However, with that, it was never called.

squeek502:
Closes #379

### DIFF
--- a/src/luv.c
+++ b/src/luv.c
@@ -660,11 +660,13 @@ LUALIB_API int luaopen_luv (lua_State* L) {
 
     lua_pushstring(L, "_loop");
     loop = (uv_loop_t*)lua_newuserdata(L, sizeof(*loop));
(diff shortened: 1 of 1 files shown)


zhaozg:
This will make main thread and normal thread loop __gc works, but threadpool not work ok. That would be solved in another PR.

PR #381, opened 2019-09-21
Optimize threadpool implementation

zhaozg:
1. reuse cached lua_State for thread pool work scheduling
2. lua_State for thread pool can be gc

### DIFF
--- a/src/work.c
+++ b/src/work.c
@@ -25,6 +25,7 @@ typedef struct {
   uv_async_t async;
   int async_cb;       /* ref, run in main, call when async message received */
   int after_work_cb;  /* ref, run in main ,call after work cb*/
+  int pool_ref;       /* ref of lua_State cache array */
 } luv_work_ctx_t;
 
 typedef struct {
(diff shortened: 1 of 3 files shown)


zhaozg:
continue https://github.com/luvit/luv/pull/380

src/work.c
```diff
@@ -1,4 +1,5 @@
 /*
+  i"close", luv_queue_close},
```
squeek502: ?

Looks like an accidental paste?

tests/test-work.lua
```diff
@@ -15,11 +15,13 @@ return require('lib/tap')(function (test)
             assert(n*n==r)
             assert(#s==4096)
             if step < count then
-                _uv.queue_work(ctx,n,s)
+                if ctx then _uv.queue_work(ctx,n,s) end
```
squeek502: Why is this change needed? Does this imply a breaking change in how `work`/`threadpool` is used?

zhaozg:
Done

tests/test-work.lua
```diff
@@ -20,6 +20,8 @@ return require('lib/tap')(function (test)
                 if (step % 100==0) then
                     p(string.format('run %d%%', math.floor(step*100/count)))
                 end
+            else
+              ctx = nil
```
squeek502: Same question here. Is this now necessary?

zhaozg: Yes, need set ctx to nil.  because ctx -> callback -> ctx loop,  if ctx not nil,  ctx will not be gc.

squeek502: You're right, and this was always true (`__gc` didn't get called in this test in older versions of `luv` too without this `ctx = nil` change)

PR #387, opened 2019-10-06
Fix os_setenv/os_unsetenv returning an error when they succeed

squeek502:
Without this patch, the added asserts in the test case would fail with:

```
  tests/test-misc.lua:127: Unknown system error 0: Unknown system error 0
  stack traceback:
        [C]: in function 'assert'
        tests/test-misc.lua:127: in function 'fn'
        ./lib/tap.lua:53: in function <./lib/tap.lua:42>
        [C]: in function 'xpcall'
        ./lib/tap.lua:42: in function 'run'
        ./lib/tap.lua:136: in function <./lib/tap.lua:113>
        [C]: at 0x00405680
not ok 18 uv.os_environ
```

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -40,6 +40,7 @@ void luv_stack_dump(lua_State* L, const char* name) {
 }
 
 static int luv_error(lua_State* L, int status) {
+  assert(status < 0);
   lua_pushnil(L);
   lua_pushfstring(L, "%s: %s", uv_err_name(status), uv_strerror(status));
   lua_pushstring(L, uv_err_name(status));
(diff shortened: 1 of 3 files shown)


src/util.c
```diff
@@ -40,6 +40,7 @@ void luv_stack_dump(lua_State* L, const char* name) {
 }
 
 static int luv_error(lua_State* L, int status) {
+  assert(status != 0);
```
SinisterRectus: assert(status < 0) ?

squeek502: That'd probably be better. Would have the potential to catch Libuv bugs where Windows error codes aren't being translated properly.

PR #388, opened 2019-10-06
Fix some tests failing when run on their own on Windows

squeek502:
`_G.isWindows` only gets set in tests/run.lua so running test-process.lua directly means `isWindows` would be nil on Windows

### DIFF
--- a/lib/utils.lua
+++ b/lib/utils.lua
@@ -12,6 +12,14 @@ else
   usecolors = false
 end
 
+if _G.jit and _G.jit.os then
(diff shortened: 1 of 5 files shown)


SinisterRectus:
Looks like `isWindows` is also used in `test-dns` and `test-signal`. This might be better placed as a property of `utils` or some sort of environmental variable set by `tap`.

squeek502:
Good call, didn't think to check where else it was used.

SinisterRectus:
LGTM.

Issue #390, opened 2019-10-07
Early initialization of userdata interferes with expected arguments


SinisterRectus:
This is a good one...

Code:
```lua
local thread = uv.new_thread()
```
Output:
```
bad argument #1 to 'new_thread' (function expected, got userdata)
```
The error message should say "got no value", but....
```c
static int luv_new_thread(lua_State* L) {
  int ret;
  size_t len;
  const char* buff;
  luv_thread_t* thread;
  int cbidx = 1;
  ...
  thread = (luv_thread_t*)lua_newuserdata(L, sizeof(*thread));
  memset(thread, 0, sizeof(*thread));
  luaL_getmetatable(L, "uv_thread");
  lua_setmetatable(L, -2);
  ...
  buff = luv_thread_dumped(L, cbidx, &len);
  ...
}
```
The call to `lua_newuserdata` pushes the `luv_thread_t` userdata to the stack in the position where the function is expected by `luv_thread_dumped`.

Obviously just pass the correct arguments to avoid the issue, but I wonder if creating the userdata early is incorrect here, or in other `uv.new_foo` calls.

squeek502:
Should definitely be handling this better. Seems like a bug.

zhaozg:
I am far away with my computer, handle it yourself or wait me some days.

PR #391, opened 2019-10-08
Fix various bugs with argument order/strictness/etc

squeek502:
Went through all instances of `luv_newuserdata` to find things similar to #390 and fixed them all.

Before:
```lua
uv.new_thread() -- bad argument #1 to 'new_thread' (function expected, got userdata)
uv.spawn("", {}) -- bad argument #3 to 'spawn' (function or callable table expected, got uv_process)
uv.new_tcp("test") -- this would coerce the string into an int
uv.new_udp() -- this would always call uv_udp_init_ex and use the uv_udp_t userdata as the arg (coerced to 0)
```

After:
```lua
uv.new_thread() -- bad argument #1 to 'new_thread' (function expected, got no value)
uv.spawn("", {}) -- no error, 3rd argument is meant to be optional
uv.new_tcp("test") -- bad argument #1 to 'new_tcp' (number expected, got string)
uv.new_udp() -- this now correctly calls uv_udp_init instead of uv_udp_init_ex, has same error as above if non-number is given as the first arg
```

### DIFF
--- a/src/process.c
+++ b/src/process.c
@@ -253,6 +253,10 @@ static int luv_spawn(lua_State* L) {
   lua_pop(L, 1);
 #endif
 
+  // this will fill the 3rd argument with nil if it doesn't exist so that
+  // the uv_process_t userdata doesn't get treated as the 3rd argument
+  lua_settop(L, 3);
+
   handle = (uv_process_t*)luv_newuserdata(L, sizeof(*handle));
   handle->type = UV_PROCESS;
   handle->data = luv_setup_handle(L, ctx);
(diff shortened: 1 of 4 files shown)


justinmk:
Yikes. Would it be easy to cover some of these with tests ?

squeek502:
I can't really think of an obvious way to test for things like this, at least with our current setup. The best thing I could think of is have a set of test scripts that call every bound function with every valid (and invalid?) set of arguments and then do some sort of validation checking of the lua stack at various points in the bound functions and/or validate the error messages (there's a tiny bit of that sort of thing [here](https://github.com/luvit/luv/blob/master/tests/test-udp.lua#L66-L73) but its just asserting that errors happen).

Luckily, in this case, these look worse than they are. The `new_udp` bug could have had unintended effects, but luckily the userdata always being coerced to 0 means that it was effectively always using `AF_UNSPEC` which is functionally identical to calling `uv_udp_init`.

PR #393, opened 2019-10-09
Make new_tcp/new_udp first argument more user-friendly/consistent

squeek502:
Allows specifying the address family as a string or an integer, consistent with how its specified in other luv bindings like getaddrinfo/getnameinfo

example:
```lua
uv.new_udp("inet")
uv.new_tcp("inet6")
```

### DIFF
--- a/src/tcp.c
+++ b/src/tcp.c
@@ -32,7 +32,21 @@ static int luv_new_tcp(lua_State* L) {
     ret = uv_tcp_init(ctx->loop, handle);
   }
   else {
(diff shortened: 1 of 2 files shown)


SinisterRectus:
Is it appropriate to use just one string here? `tcp_bind` and `udp_bind` accept a table of `name = boolean` fields for their flags, but I'm not familiar enough with the API right now to know if they use the same concept.

squeek502:
They are different. Will expand on this more in #399

PR #344, opened 2019-06-23
Test with LuaRocks 3.1.3

blueyed:
Debugging issues with LuaRocks upgrade with Neovim (https://github.com/neovim/neovim/pull/10292).

PR #396, opened 2019-10-09
Remove MSVC 2013 jobs from appveyor CI

squeek502:
This should cut the time it takes to run Appveyor CI in half (from ~50 minutes to ~25 minutes) and MSVC 2013 is old enough to stop running CI for I think (and even 2015 might be worth changing to 2017?).

EDIT: Related: https://github.com/luvit/luv/pull/344#issuecomment-515631371

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,10 @@
 os: Visual Studio 2015
 
-# Test with the latest two releases of MSVC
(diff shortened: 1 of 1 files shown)


zhaozg:
I prefer 2017

squeek502:
Will update to 2017 in a different PR.

Issue #400, opened 2019-10-09
Error handling API


SinisterRectus:
Libuv's [error handling API](http://docs.libuv.org/en/v1.x/errors.html) is not exposed in the luv module. Internal errors are instead returned as an assertable `nil, err, name`. This is fine, but there are issues related to this:

1. Returning `nil` instead of a negative integer is not consistent with the fact that some other luv functions return `0` on success. While this may be normal in C, I don't think Lua users expect this. Should we change these returns to `true`? Would this be a breaking change?

2. Some luv functions return `nil` because their internal libuv functions return `void`. How do we indicate that these functions have not failed? Do we just have to document the error signature *must* match `nil, err, ...` and that you cannot reliably `assert` all luv calls?

squeek502:
1. `0` is truthy in Lua (`if 0 then print('true') end`) so changing those returns to true should be fairly safe (unless people are doing `== 0` checks).
2. Consistently returning `true` on success would be nice

SinisterRectus:
Luv error behavior was never documented, so maybe we're allowed to change it? Seems sketchy though. I worry that somewhere, someone is checking for `0`.

There are [places](<https://github.com/luvit/lit/blob/67e7219314589f1cfbf1347cfdc9496dae3dd97d/deps/coro-channel.lua#L133-L136>) where we use the implicit check. I cannot find any explicit `0` checks in luvit or lit (except for process exit codes, which I think is a different thing).

----

On a related note, Lua 5.4 introduces the concept of `fail` for function returns.

> The notation **fail** means a return value representing some kind of failure or the absence of a better value to return. Currently, **fail** is equal to **nil**, but that may change in future versions. The recommendation is to always test the success of these functions with `(not status)`, instead of `(status == nil)`.

For example:

from `tonumber` in Lua 5.3:
> If the string is not a valid numeral in the given base, the function returns **nil**.

from `tonumber` in Lua 5.4:
> If the string is not a valid numeral in the given base, the function returns **fail**.

squeek502:
I'd be somewhat surprised if anyone is doing `== 0` checks. But, since 0 is truthy anyway, maybe it's just not worth changing, even if the risk is minimal?

SinisterRectus:
It looks like our API is inconsistent. Most functions that don't return an otherwise useful value return `0` to indicate success, but some return a boolean or nothing at all.

Functions that coerce the status int to a boolean:
- `uv.run`
- `uv.thread_join`
- `uv.os_setenv`
[...]

PR #403, opened 2019-10-09
Standardize/factor out common result handling code

squeek502:
This is a preparatory step that will make #400 much easier but is probably a good idea regardless of what we end up deciding to do there.

### DIFF
--- a/src/async.c
+++ b/src/async.c
@@ -58,7 +58,5 @@ static int luv_async_send(lua_State* L) {
 
   luv_thread_arg_set(L, arg, 2, lua_gettop(L), 0);
   ret = uv_async_send(handle);
-  if (ret < 0) return luv_error(L, ret);
-  lua_pushinteger(L, ret);
-  return 1;
+  return luv_result(L, ret);
 }
--- a/src/check.c
+++ b/src/check.c
@@ -45,16 +45,12 @@ static int luv_check_start(lua_State* L) {
   int ret;
   luv_check_callback(L, (luv_handle_t *)handle->data, LUV_CHECK, 2);
   ret = uv_check_start(handle, luv_check_cb);
-  if (ret < 0) return luv_error(L, ret);
-  lua_pushinteger(L, ret);
-  return 1;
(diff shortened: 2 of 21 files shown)


SinisterRectus:
The only issue I see here is that some of these functions return an integer as part of their normal operation. If we decide to change the behavior for returning success status, we wouldn't want to include these functions.

squeek502:
> some of these functions return an integer as part of their normal operation

Could you provide an example of one? I'll double check, but I thought the pattern I replaced wouldn't include any of those.

SinisterRectus:
The ones I can find are:
- `send_buffer_size` (if not setting the size)
- `recv_buffer_size` (if not setting the size)
- `backend_fd`
- `try_write`

I notice also that `backend_timeout` is a function that can return -1 successfully. That is already set up to not push an error though.

squeek502:
`send_buffer_size`/`recv_buffer_size` are bugged: https://github.com/luvit/luv/issues/404

squeek502:
* `send_buffer_size`: fixed by #405, now only uses luv_result in the setter case
* `recv_buffer_size`: fixed by #405, now only uses luv_result in the setter case
* `backend_fd`: this is also bound wrong--it never returns a UV_E error code, it just returns -1 when unsupported (on Windows). It's now unchanged in this PR, will address it in a separate PR
* `try_write`: Updated this to use better variable names so that it doesn't get mistaken for the luv_result pattern

Will double check the rest of the changes in this PR to make sure that there are no other functions that shouldn't be using `luv_result` before merging.

squeek502:
Ok, I'm now confident that this only affects functions that only return their success/error state. Will squash/merge if there are no other comments.

SinisterRectus:
I'll make one more pass through and give it an okay later.

SinisterRectus:
Can you check that `fileno` is operating properly? Other than that, looks good to go.

squeek502:
Made a separate issue for `fileno` (#408) since it definitely should not be using `luv_result`.

PR #407, opened 2019-10-11
Update Appveyor CI to use MSVC 2017/2019

squeek502:
- Seperates the jobs into two distinct types:
  * `FROM_DEPS`: Uses 2015 / 2017 / 2019 on both x86 and x64; compiles from git using msvcbuild.bat and runs all the tests
  * `LUAROCKS`: Uses only MSVC 2017 x64; runs LuaRocks builds only, using both LuaJIT 2.1 and Lua 5.3
- Updates msvcbuild.bat to use MSVC 2017 by default
- Full set of jobs completes in around 30 minutes, similar to what we have now
- Example build of this branch: https://ci.appveyor.com/project/squeek502/luv/builds/28038711

Left my commits messy, will squash them before merging.


squeek502:
In doing this, I noticed that we're not running the tests on Windows with PUC Lua 5.3 (currently or after this PR). That's probably something that should happen.

Issue #399, opened 2019-10-09
Some minor API inconsistencies


SinisterRectus:
I'm working on a much needed update to the luv docs. It's nearly done, but I want to highlight some API inconsistencies that I found. None of these seem to be a big deal, some of them are nitpicky, but I wanted to get this list here while I have it

----

- `uv.tty_set_mode` accepts raw `tty_set_mode_t` enum instead of a lua string
- `uv.getaddrinfo` `hints` take raw integer values (I think they are in `uv.constants`?)

compare to:

- `uv.run` mode accepts lua string instead of raw `uv_run_mode` enum
- `uv.udp_set_membership` accepts lua string instead of raw `uv_membership` enum
- `uv.poll_start` accepts lua string instead of raw `int` for events

`uv.tty_set_mode` could optionally accept a string so that the change isn't breaking. Should we allow other enum-accepting functions to accept ints? Should we expose more enums at `uv.constants`?

----

- `uv.new_tcp` accepts raw `int` for flags
- `uv.new_udp` accepts raw `int` for flags

compare to:

- `uv.tcp_bind` accepts a table of booleans instead of raw `int` for flags
- `uv.udp_bind` accepts a table of booleans instead of raw `int` for flags
- `uv.fs_event_start` accepts a table of booleans instead of raw `int` for flags

Might be fixed? See #393 and #398.

----

- `uv.getaddrinfo`returns table with `addr`, `port`, `family`

compare to:

- `uv.getnameinfo` accepts table with `ip`, `port`, `family`
- `uv.tcp_getpeername` returns table with `ip`, `port`, `family`
- `uv.tcp_getsockname` returns table with `ip`, `port`, `family`
- `uv.udp_getsockname` returns table with `ip`, `port`, `family`
- `uv.udp_getpeername` returns table with `ip`, `port`, `family`

You can't directly use a `getaddrinfo` result in `getnameinfo`, but would you want to?

----

That's all I have for here. I'll be opening other issues or maybe PRs for things that I think deserve there own.

squeek502:
Replying to your comment in #393 here:

The most future proof way to bind `new_udp`/`new_tcp` would be to accept a table with key `"family"` that can either be a string like `"inet"` or a raw AF_ enum integer (this matches `luv_getaddrinfo`), because techincally the `flags` param of `uv_new_udp` could have more added to it in the future (right now it uses `flags & 0xFF` for the `AF_` enum part).

[...]

PR #409, opened 2019-10-11
Update and expand docs

### LINKED ISSUE #118: Finish porting libuv docs

### LINKED ISSUE #144: Cleanup and document uv_work_t and uv_thread_t bindings.

### LINKED ISSUE #378: update docs for new api.
> https://github.com/luvit/luv/blob/master/docs.md

SinisterRectus:
First pass at improving docs.md.

Closes #118
Closes #144
Closes #378

This is a total rewrite so the diff might be overwhelming. Sorry that it's all one big commit. Since the previous version, I cleaned up or expanded the existing sections and added:
- Error handling
- Version checking
- uv_fs_event_t
- uv_fs_poll_t
- File system operations
- Thread pool work scheduling
- DNS utility functions
- Threading and synchronization utilities
- Miscellaneous utilities

TODO
- [x] `new_udp` and `new_tcp` flags - Documented.
  - ~~New change; haven't had a chance to look at it.~~ 
- [x] `translate_sys_error` - Ignored.
  - ~~This is the first error function exposed to Lua, so I'm not sure what this is doing here.~~
- [x] `uv_req_t` - Documented.
  - ~~We only expose `uv_cancel` but we might want to document which subtypes are `uv_req_t` and which ones can be canceled.~~ (See #416)
- [x] `tcp_write_queue_size` - Deprecated.
  - ~~This isn't a libuv function, so I'm not sure what it's doing.~~
- [x] `tcp_close_reset` - Documented.
  - ~~Returned values are unclear to me.~~
- [x] `fs_readdir` - Documented.
  - ~~Behavior compared to `fs_scandir_next` is unclear to me.~~
- [x] Parameter types (maybe)  - Deferred
  - ~~I documented return types, but not parameters. This is going to be a bit of work, so I may save it for another time.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ This will build luv as a module library. Module libraries are plugins that are
 not linked into other targets.
 
 #### Build with PUC Lua 5.3
-By default luv is linked with LuaJIT 2.0.4. If you rather like to link luv
+By default luv is linked with LuaJIT 2.1.0-beta3. If you rather like to link luv
 with PUC Lua 5.3 you can run make with:
 
 ```
--- a/docs.md
+++ b/docs.md
@@ -247,7 +247,13 @@ end)
 
 [`uv_req_t`]: #uv_req_t--request-handle
 
-TODO
+`uv_req_t` is the base type for all libuv request types.
+
+### `uv.cancel(req)`
+
+Cancel a pending request. Fails if the request is executing or has finished
+executing. Only cancellation of `uv_fs_t`, `uv_getaddrinfo_t`,
+`uv_getnameinfo_t` and `uv_work_t` requests is currently supported.
 
 ## `uv_handle_t` — Base handle
 
--- a/docs.md
+++ b/docs.md
@@ -1731,13 +1731,13 @@ operations.
 
 All file descriptors `fd` are integers.
 
-All functions take an optional `callback`. If the callback is omitted, the
-request is completed synchronously with values returned to the original caller.
-If a callback is provided, the request `uv_fs_t` is returned to the original
-caller and is asynchronously completed with results passed to the callback.
+If a `callback` is passed to functions that accept one, the request `uv_fs_t` is
+returned to the original caller and is asynchronously completed with results
+passed to the callback. If omitted, the request is completed synchronously with
+values returned to the original caller.
 
-All `mode`s are integers indicating access permissions. To use an octal
-representation, pass an explicit `tonumber` call such as `tonumber("0666", 8)`.
+All `mode`s are integers indicating access permissions in decimal format. To use
+an octal format, use a `tonumber`. For example: `tonumber("0666", 8)`.
 
 ### `uv.fs_close(fd, [callback])`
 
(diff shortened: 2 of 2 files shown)


squeek502:
Comments on the TODOs:

- `new_udp` and `new_tcp`: see https://github.com/luvit/luv/issues/399#issuecomment-540185444
- `translate_sys_error`: This was made public in Libuv in 1.10.0, and is the only public Libuv error function, which explains why its the only one bound. How its bound is interesting, it returns the `err, name` of the `fail` tuple, so I guess it would enable constructing a `fail` return from Lua (`return nil, uv.translate_sys_error(SOME_ERR_CODE_THAT_COMES_FROM_CALLING_A_SYSTEM_FUNCTION)`)
- `tcp_write_queue_size`: Looks like `uv_stream_get_write_queue_size` got added in 1.19.0 but Luv never bound it. I'll create a new issue for all unbound functions that remain.
- `tcp_close_reset`: Looks like a bug, it returns whatever argument happens to be at the top of the stack (usually the callback or the uv_tcp_t)
- `fs_readdir`: It is non-blocking whereas `fs_scandir_next` is blocking. Example:

```lua
local uv = require('luv')

local check = uv.new_check()
uv.check_start(check, function()
  print('loop iteration')
end)

uv.fs_scandir('.', function(err, req)
  assert(not err, err)
  while true do
    local name, ftype = uv.fs_scandir_next(req)
    if not name then break end
    print(name, ftype)
  end
  uv.check_stop(check)
end)
```
prints:
```
loop iteration
.ci     directory
.editorconfig   file
.git    directory
.gitignore      file
.gitmodules     file
...
```

whereas:
```lua
local uv = require('luv')
local dump = require('lib/utils').dump

local check = uv.new_check()
uv.check_start(check, function()
  print('loop iteration')
end)

uv.fs_opendir('.', function(err, dir)
  assert(not err, err)

  local function readdir_cb(err, files)
    assert(not err, err)
    if files then
      print(dump(files))
      uv.fs_readdir(dir, readdir_cb)
    else
      assert(uv.fs_closedir(dir))
      uv.check_stop(check)
    end
  end

  uv.fs_readdir(dir, readdir_cb)
end)

uv.run()
```

prints:
```
loop iteration
loop iteration
{ { name = ".ci", type = "directory" } }
loop iteration
loop iteration
{ { name = ".editorconfig", type = "file" } }
loop iteration
loop iteration
{ { name = ".git", type = "directory" } }
loop iteration
loop iteration
{ { name = ".gitignore", type = "file" } }
loop iteration
{ { name = ".gitmodules", type = "file" } }
loop iteration
```

squeek502:
I was wrong about `uv_translate_sys_error` being the only public Libuv error function. All the [functions here](http://docs.libuv.org/en/v1.x/errors.html#api) as just as public.

SinisterRectus:
Most of the remaining sections are now done.

I am going to ignore `translate_sys_error` for now, as #400 and #410 are still being considered.

I added deprecation notices for `uv.getpid` and `uv.tcp_write_queue_size`.

I am going to defer documenting parameter types until a later date, if done at all. With the parameter names and function description, there is usually enough information to figure out the type (as in the Lua reference manual), but sometimes a table is nice at a glance:

> ### `uv.timer_start(timer, timeout, repeat, callback)`
> | Parameter | Type |
> | - | - |
> | timer | uv_timer_t |
> | timeout | integer |
> | repeat | integer |
> | callback | function |
> 
> If `timeout` is zero, the callback fires on the next event loop iteration. If
> `repeat` is non-zero, the callback fires first after `timeout` milliseconds and
> then repeatedly after `repeat` milliseconds.

zhaozg:
Thanks you nice job.

PR #194, opened 2015-11-18
gc: setup a proper gc for the uv_loop

rphillips:
fixes: #193

Issue #193, opened 2015-11-17
Random segfault when running a busted test suite that calls luv


tarruda:
Hi

I'm refactoring Neovim [lua-client](https://github.com/neovim/lua-client) to use luv as the event loop backend. Currently it uses a simple C module that talks to libuv directly: https://github.com/neovim/lua-client/blob/master/nvim/loop.c

In a [PR](https://github.com/neovim/lua-client/pull/20) I have replaced the C module by a pure lua version and uses luv to talk to libuv: https://github.com/tarruda/lua-client/blob/refactor-loop-to-use-luv/nvim/loop.lua

The lua client tests currently pass in the luv branch, the only issue is that I get random segfaults after the test suite ends. Here's a backtrace taken from the core dump(linux x64):

```
[New LWP 27870]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/home/tarruda/dev/lua-client/.deps/usr/bin/lua -e package.path="/home/tarruda/d'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00002b02fb1bd810 in ?? ()
#0  0x00002b02fb1bd810 in ?? ()
#1  <signal handler called>
#2  _int_free (av=0x2b02fa951760 <main_arena>, p=0x11cfc00, have_lock=0) at malloc.c:3918
#3  0x000000000041a8ac in l_alloc (ud=0x0, ptr=0x11cfc10, osize=24, nsize=0) at lauxlib.c:631
#4  0x000000000040da1b in luaM_realloc_ (L=0x11ca010, block=0x11cfc10, osize=24, nsize=0) at lmem.c:79
#5  0x000000000040b780 in luaF_freeproto (L=0x11ca010, f=0x11d9060) at lfunc.c:147
#6  0x000000000040c9db in freeobj (L=0x11ca010, o=0x11d9060) at lgc.c:380
#7  0x000000000040cb75 in sweeplist (L=0x11ca010, p=0x11ca0f0, count=18446744073709541977) at lgc.c:424
#8  0x000000000040cebe in luaC_freeall (L=0x11ca010) at lgc.c:487
#9  0x0000000000412208 in close_state (L=0x11ca010) at lstate.c:108
#10 0x0000000000412718 in lua_close (L=0x11ca010) at lstate.c:212
#11 0x0000000000404d7e in main (argc=12, argv=0x7ffc5c685608) at lua.c:389
[...]

PR #414, opened 2019-10-14
Fix uv.shutdown cb being called when canceled

squeek502:
This could happen during lua_close, and after absolutely all handles have been closed, so there's no point in running the callback.

I'm not sure, but there *might* be other legitimate situations where UV_ECANCELED should be passed along to the callback. If so, then there would need to be a more involved fix here to handle when the stream is closed via lua_close/__gc

---

This is something I found [when looking at our CI's `PROCESS_CLEANUP_TEST` jobs](https://travis-ci.org/luvit/luv/jobs/597430935) (see https://github.com/luvit/luv/pull/193 and https://github.com/luvit/luv/pull/194 for when/why they exist) and noticing that they were throwing a Lua error (and somehow not failing the CI?). I think this has been broken for a long time (likely ever since `uv_handle_t` got a `__gc` that closed itself).

This is what happens when running `./tests/test-sigchld-after-lua_close.sh`

```
{ pid = 5757, handle = uv_process_t: 0xbdfed0 }
Uncaught Error: examples/talking-to-children.lua:28: bad argument #1 to 'close' (Expected uv_handle userdata)
stack traceback:
	[C]: in function 'luv.close'
	examples/talking-to-children.lua:28: in function <examples/talking-to-children.lua:27>
	[C]: in ?
```
This is not caused by not passing in a `uv_handle` userdata, but rather because the `uv_handle` has been closed, and `onshutdown` is being called during `lua_close`. Adding a print statement to the `onshutdown` callback gives you this:

```
Uncaught Error: ./lib/utils.lua:155: bad argument #1 to 'tostring' (Expected uv_handle userdata)
stack traceback:
        [C]: in function 'tostring'
        ./lib/utils.lua:155: in function 'print'
        examples/talking-to-children.lua:28: in function <examples/talking-to-children.lua:27>
```

i.e. the `__tostring` metamethod of the userdata is failing because the userdata is a closed (via __gc) uv_stream_t.

### DIFF
--- a/src/stream.c
+++ b/src/stream.c
@@ -42,8 +42,12 @@ static uv_stream_t* luv_check_stream(lua_State* L, int index) {
 static void luv_shutdown_cb(uv_shutdown_t* req, int status) {
   luv_req_t* data = (luv_req_t*)req->data;
   lua_State* L = data->ctx->L;
-  luv_status(L, status);
-  luv_fulfill_req(L, (luv_req_t*)req->data, 1);
+  // if the stream was closed (e.g. __gc'd) then the shutdown gets canceled
+  // and the callback shouldn't be run
+  if (status != UV_ECANCELED) {
+    luv_status(L, status);
+    luv_fulfill_req(L, (luv_req_t*)req->data, 1);
+  }
   luv_cleanup_req(L, (luv_req_t*)req->data);
   req->data = NULL;
 }


squeek502:
Some debug printing to better explain what's happening:

```
luv_shutdown called with handle: 000000000035FF50 -> req: 00000000001E8DE0
handle gc | 00000000003601D0 -> not closed, closing
handle gc | 000000000035FF50 -> not closed, closing
loop gc -> uv_walk -> uv_run
uv_process_endgames -> uv_pipe_endgame -> req->cb(req, UV_ECANCELED);
luv_shutdown_cb for req: 00000000001E8DE0
cb called    ECANCELED
Uncaught Error: bad argument #1 to '?' (Expected uv_handle userdata)
stack traceback:
        [C]: at 0x07fee3d32da0
        [C]: in function 'print'
        test.lua:24: in function <test.lua:22>
        [C]: at 0x07fee3d49950
```

Minimal reproduction:

```lua
local uv = require('luv')

local stdin = uv.new_pipe(false)

local process, pid = uv.spawn(jit.os == "Windows" and "cmd.exe" or "cat", {
  stdio = {stdin, nil, nil}
})

assert(uv.shutdown(stdin, function(...)
	print("cb called", ...)
	print(process)
end))
```

squeek502:
Worth noting that doing this from the Lua side also fixes this particular case:

```
uv.shutdown(stdin, function(err)
  if err == "ECANCELED" then
    return
  end
  print(process)
end)
```

So instead of doing this on the C side we could pass ECANCELED to the callback (even during lua_close) and it would be up to the user to handle that case in their callback.

zhaozg:
> So instead of doing this on the C side we could call the callback with ECANCELED (even during lua_close) and it would be up to the user to handle that case in their callback.

What is prefer solution, lua side or c side?

squeek502:
I think it would be better on the C side if we can be sure that `UV_ECANCELED` is only possible during `lua_close`, but I'm not sure if that's true. I'll do some more testing.

zhaozg:
https://github.com/libuv/libuv/search?q=UV_ECANCELED&unscoped_q=UV_ECANCELED shows some api will cause UV_ECANCELED, e.g. uv_cancel()

squeek502:
I meant `UV_ECANCELED` for `shutdown`. Herr are all the ways in Libuv that the shutdown cb can be called with `UV_ECANCELED`:

- https://github.com/libuv/libuv/blob/f9c35197e64202099b691d53df7ea7db480b390c/src/unix/stream.c#L471-L480
- https://github.com/libuv/libuv/blob/f9c35197e64202099b691d53df7ea7db480b390c/src/win/tty.c#L2221-L2222
- https://github.com/libuv/libuv/blob/f9c35197e64202099b691d53df7ea7db480b390c/src/win/pipe.c#L367-L377
- https://github.com/libuv/libuv/blob/958e85fbd842be06da8484305996f42149ba5039/src/win/tcp.c#L213-L234

Relevant Libuv test:
- https://github.com/libuv/libuv/blob/e4087dedf837f415056a45a838f639a3d9dc3ced/test/test-shutdown-close.c

zhaozg:
this should be merge now?

squeek502:
Don't want to merge it yet, still not sure if handling this on the C side is correct and want to do some more testing.

Take a look at https://github.com/luvit/luv/issues/433, it's the more important thing to get fixed first.

squeek502:
Going to focus on this today.

squeek502:
This is not the right fix. The shutdown callback can be called with `UV_ECANCELED` with the following code:

```lua
local uv = require('luv')
local isWindows = package.config:find("\\") and true or false

local stdin = uv.new_pipe(false)

handle, pid = uv.spawn(isWindows and "cmd.exe" or "cat", {
  stdio = {stdin, nil, nil}
}, function(...)
  print("onexit", ...)
end)

assert(uv.shutdown(stdin, function(...)
  print("onshutdown", ...)
end))
uv.close(stdin, function(...)
  print("onclose", ...)
end)

uv.run()
```

PR #420, opened 2019-10-27
Remove c-api directory from #include for compat-5.3.h

jamessan:
This better matches how it would be found when building against system-installed lua-compat53.

---
I realize there's not a _ton_ of precedence here, since a lot of packages simply vendor lua-compat53, but Debian's [codesearch](https://codesearch.debian.net/search?q=compat-5%5C.3%5C.h+-pkg%3Alua-compat53+-path%3A.*compat-5%5C.3%5C.c&literal=1) shows a general trend.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ script:
   - mkdir build/lib
   - cp build/deps/libuv/libuv_a.a build/lib/libuv.a
   - cp -a deps/libuv/include build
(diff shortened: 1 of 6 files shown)


blueyed:
Note that this has been (mostly) reverted in 60946d3 (https://github.com/luvit/luv/pull/424#issuecomment-727896218).

Issue #382, opened 2019-09-24
More memory leaks found with address sanitizer


vanc:
This can be considered a continuation from issue #379. I didn't report all issues as I didn't know #379 was resolved so quickly. Thank you guys!

The first one was related to signal handling. It was from test-signal.lua.

=================================================================
==6844==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 128 byte(s) in 1 object(s) allocated from:
    #0 0x4e8196 in realloc /home/llvm/llvm/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:165
    #1 0x7f14b2dbbbe0 in maybe_resize /home/lua-projects/libuv/src/unix/core.c:850:14
    #2 0x7f14b2dbbbe0 in uv__io_start /home/lua-projects/libuv/src/unix/core.c:889
    #3 0x7f14b2dded06 in uv__signal_loop_once_init /home/lua-projects/libuv/src/unix/signal.c:273:3
    #4 0x7f14b2dded06 in uv_signal_init /home/lua-projects/libuv/src/unix/signal.c:320
    #5 0x7f14b2dd4af2 in uv_loop_init /home/lua-projects/libuv/src/unix/loop.c:72:9
    #6 0x7f14b2d82b3f in luaopen_luv /home/lua-projects/luv/luv.c:674:11
    #7 0x5cb016 in lj_BC_FUNCC /home/lua-projects/luajit/src/lj_vm.S:809

SUMMARY: AddressSanitizer: 128 byte(s) leaked in 1 allocation(s).
  "exit"	{ signal = 0, code = 1, pid = 6844 }
Uncaught Error: ./tests/test-signal.lua:25: assertion failed!

After finish all tests, more leaks were dumped out.

ok 98 work - test threadpool
#2 failed tests

=================================================================
==6710==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 312 byte(s) in 1 object(s) allocated from:
    #0 0x4e7d07 in malloc /home/llvm/llvm/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:146
    #1 0x7f52a138d084 in luv_newuserdata /home/lua-projects/lua-modules/luv/src/handle.c:20:18
    #2 0x7f52a138d084 in luv_new_tty /home/lua-projects/lua-modules/luv/src/tty.c:32
    #3 0x5cb016 in lj_BC_FUNCC /home/lua-projects/luajit/src/lj_vm.S:809

[...]

PR #422, opened 2019-10-28
tap: Make os.exit call lua_close on Lua versions that support it

squeek502:
Second optional argument to os.exit is a boolean that determines if lua_close should be called while exiting. This was added in PUC Lua 5.2 and backported to LuaJIT.

This allows the tests to provide accurate memory leak information even with failing tests (memory leak info will still be inaccurate with failing tests on PUC Lua 5.1, since PUC Lua 5.1 os.exit bypasses lua_close).

Thanks to @vanc for the fix, see https://github.com/luvit/luv/issues/382#issuecomment-546032911

### DIFF
--- a/lib/tap.lua
+++ b/lib/tap.lua
@@ -105,7 +105,7 @@ local function run()
   uv.run()
 
   if failed ~= 0 then
-    os.exit(-failed)
+    os.exit(-failed, true)
   end
 end
 


vanc:
Thanks @squeek502. I haven't yet created any PR on github. Will learn that later to save you guys some time.

Thank you for the excellent work!

PR #424, opened 2019-10-28
fixup deps lua-compat-5.3 include

zhaozg:


squeek502:
What does this fix?

zhaozg:
> What does this fix?

Think about this when I give LUA_COMPAT53_DIR by env variable,   the old code will not work
https://github.com/luvit/luv/blob/master/CMakeLists.txt#L155-L158

appveyor.yml
```diff
@@ -82,6 +82,6 @@ for:
   - mkdir build\lib
   - cp build.luarocks\deps\libuv\Release\uv_a.lib build\lib\uv.lib
   - cp -a deps\libuv\include build
-  - ps: luarocks make rockspecs\$(ls rockspecs) LIBUV_DIR=build LUA_COMPAT53_INCDIR=deps/lua-compat-5.3/c-api CFLAGS="/nologo /MT /O2"
+  - ps: luarocks make rockspecs\$(ls rockspecs) LIBUV_DIR=build LUA_COMPAT53_INCDIR=deps/lua-compat-5.3 CFLAGS="/nologo /MT /O2"
```
squeek502: This should probably still be `LUA_COMPAT53_INCDIR=deps/lua-compat-5.3/c-api`.

`LUA_COMPAT53_INCDIR` passed to LuaRocks is the folder that contains the `.h` file itself. `LUA_COMPAT53_DIR` passed to CMake is the `lua-compat-5.3` folder with `c-api` inside it.

If your goal is full backwards compatibility, I think we might need to revert #420 completely.

zhaozg: My mistake, I'll try to fix.

SinisterRectus:
Is the change to pipe.c intentional? I purposefully did not assign 0 to ipc because I thought it should be 0 by default on declaration; same reason why I didn't coerce nil to 0.

zhaozg:
> Is the change to pipe.c intentional? I purposefully did not assign 0 to ipc because I thought it should be 0 by default on declaration; same reason why I didn't coerce nil to 0.

Hi @SinisterRectus please look at [here](https://stackoverflow.com/questions/1597405/what-happens-to-a-declared-uninitialized-variable-in-c-does-it-have-a-value), and never assume that will be 0

SinisterRectus:
Oh okay. It worked when I tested it, but if it's undefined behavior, then a change is good. In that case, I would consider using:
```c
  int ipc, ret;
  luv_ctx_t* ctx = luv_context(L);
  if (lua_isboolean(L, 1) || lua_isnil(L, 1)) {
    ipc = lua_toboolean(L, 1);
  } else {
    luaL_argerror(L, 1, "Expected boolean or nil");
  }
```
With the explicit check for nil, the 0 assignment and argcheck are redundant. (I don't know why Lua doesn't have `luaL_checkboolean` for this).

zhaozg:
compile is not smart enough,so assign 0 is safe

SinisterRectus:
Then just remove the argcheck? Which reminds me of another point:

```c
(lua_isboolean(L, 1) || lua_isnil(L, 1))
```
should be
```c
(lua_isboolean(L, 1) || lua_isnoneornil(L, 1))
```
if we want `uv.new_pipe()` to work.

zhaozg:
i think only need lua_toboolean

SinisterRectus:
Yes, that should also work, if the argcheck is not removed.

```c
  int ipc, ret;
  ipc = 0;
  luv_ctx_t* ctx = luv_context(L);
  if (lua_isboolean(L, 1)) {
    ipc = lua_toboolean(L, 1);
  } else {
    luaL_argcheck(L, lua_isnoneornil(L, 1), 1, "Expected boolean or nil");
  }
```

SinisterRectus:
Test:
```lua
assert(uv.new_pipe(true))
assert(uv.new_pipe(false))
assert(uv.new_pipe(nil))
assert(uv.new_pipe())
assert(not pcall(uv.new_pipe, 'foo'))
```

zhaozg:
how about
```lua
  int ipc, ret;
  luv_ctx_t* ctx = luv_context(L);
  ipc = lua_tostring(L, 1);
```
In lua world, only nil,false,LUA_NONE are false

SinisterRectus:
Hmmm. [`toboolean`](https://www.lua.org/manual/5.3/manual.html#lua_toboolean) coerces other values, so `uv.new_pipe('foo')` would be equivalent to `uv.new_pipe(true)`. I think we should avoid that.

zhaozg:
@SinisterRectus please review https://github.com/luvit/luv/pull/427

blueyed:
> > What does this fix?
> 
> Think about this when I give LUA_COMPAT53_DIR by env variable, the old code will not work
> https://github.com/luvit/luv/blob/master/CMakeLists.txt#L155-L158
> LUA_COMPAT53_DIR should point to deps/lua-compat-5.3, not include c-api

Code ref: https://github.com/luvit/luv/blob/b461b6fa384c5722b8622a49eb045c52ea3810d5/CMakeLists.txt#L155-L157

This appears to revert dec4661 incompletely, since the "c-api" prefix is still required with luarocks:
https://github.com/luvit/luv/blob/464b7350487da5c2153cd60b44ab30ec3d8df759/.travis.yml#L40
https://github.com/luvit/luv/blob/464b7350487da5c2153cd60b44ab30ec3d8df759/rockspecs/luv-scm-0.rockspec#L29-L41
/cc @jamessan

blueyed:
I think it is fine currently - I was also confused by `LUA_COMPAT53_INCDIR` (rockspec) vs `LUA_COMPAT53_DIR`.

> LUA_COMPAT53_DIR should point to deps/lua-compat-5.3, not include c-api

Ok.
`LUA_COMPAT53_INCDIR` then is meant to point to the "include" directory, i.e. "c-api".
It is unfortunate that this requires changes to building the luarock, but I guess that is fine then for now - given that it has been released since a while already.

Issue #397, opened 2019-10-09
Audit string coercion throughout luv


squeek502:
From https://github.com/luvit/luv/issues/127#issuecomment-449615737

> Another unrelated thing (which probably requires a separate issue to be opened) is that one must be very careful when converting stack values to strings via any of `lua_tostring`, `luaL_checkstring`, `luaL_optstring` and other functions that may cause coercion. For example, I can see that in your code when a table of strings is written to a stream, there is no protection that an element might in fact be a number. When a `uv_buf_t` vector is being initialized, the item is converted to a string on the stack and its pointer is saved in the vector. All seems good. However, after the conversion is made and before the actual string is written to a socket in a `uv_write_t` request, garbage collection may happen effectively invalidating the saved pointer because its source string existed briefly on the stack as a result of coercion (the item in the table is still a number, not a string). And that means in turn that undefined data will actually be written.

> Btw, the table that is used to form a `uv_buf_t` vector may be altered even **after** the call directly in Lua which will lead to the same outcome. There is no protection against it.

> I haven't checked through your code much, but there are many other caveats when using Lua C API that one must be aware of. They are all more or less mentioned in the manual though.

This is something I ran into when writing https://github.com/luvit/luv/pull/393 (if `lua_isstring` was the first if check, then a number passed as the argument would return true from `lua_isstring` and get coerced into a string) and it's something that we need to be more careful of and are probably using incorrectly.

squeek502:
Confirmed that the `uv_buf_t` problem mentioned in the quote above is accurate and reproducible via:

```lua

local uv = require('luv')

local path = "_test_"
local fd = assert(uv.fs_open(path, "w", tonumber("0666", 8)))
print("writing")
do
  local t = {"with\n", 600, "lines\n"}
  uv.fs_write(fd, t, -1, function()
    print("written")
    uv.fs_close(fd)
  end)
end
local count = collectgarbage("count")
collectgarbage("collect")
print("garbage collected", count - collectgarbage("count"))

uv.run()
```

Note that it's the coerced number that is failing; I couldn't get a crash when using strings (probably due to how strings are handled internally by Lua). When run the above with PUC Lua 5.1 and Valgrind I get:

```
==13484== Memcheck, a memory error detector
==13484== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==13484== Using Valgrind-3.15.0 and LibVEX; rerun with -h for copyright info
==13484== Command: lua test.lua
==13484==
writing
[...]

PR #426, opened 2019-10-28
Stop bindings that use uv_buf_t coercing numbers to strings

squeek502:
When a number is coerced to a string, that string is temporary and could be garbage collected before the uv_buf_t is used, leading to use-after-free.

Part of #397, see https://github.com/luvit/luv/issues/397#issuecomment-546791557 for the reproduction test case.

Also improves the error messages a bit, they now look like this:
```
bad argument #2 to 'fs_write' (data must be string or table of strings, got number)
bad argument #2 to 'fs_write' (expected table of strings, found number in the table)
```

---

Note that this same pattern of use-after-free could also affect how we pass strings to `uv_buf_t`. From the Lua reference manual: 

> Because Lua has garbage collection, there is no guarantee that the pointer returned by lua_tolstring will be valid after the corresponding value is removed from the stack.

I haven't found a way to reproduce this with strings yet (the memory layout of the internal Lua string table makes it less likely I believe), but it's probably possible.

### DIFF
--- a/src/fs.c
+++ b/src/fs.c
@@ -466,12 +466,12 @@ static int luv_fs_write(lua_State* L) {
     bufs = luv_prep_bufs(L, 2, &count);
     buf.base = NULL;
   }
-  else if (lua_isstring(L, 2)) {
-    luv_check_buf(L, 2, &buf);
+  else if (luv_isstring_strict(L, 2)) {
+    luv_prep_buf(L, 2, &buf);
     count = 1;
   }
   else {
-    return luaL_argerror(L, 2, "data must be string or table of strings");
+    return luaL_argerror(L, 2, lua_pushfstring(L, "data must be string or table of strings, got %s", luaL_typename(L, 2)));
   }
 
   offset = luaL_checkinteger(L, 3);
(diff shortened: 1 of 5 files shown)


SinisterRectus:
Would it be too much work to pre-coerce numbers to strings and keep the references alive until after the write completes?

squeek502:
Is there a use case for allowing numbers here? In my mind passing numbers to one of these `write` functions is probably a bug, and the coercion was a bug as well. Is your intention keeping backwards compatibility here just in case?

Beyond that, though, keeping the references alive is a fix that needs to happen for the strings as well just to be safe. This actually happens in `luv_write`/`luv_write2`:
https://github.com/luvit/luv/blob/0ea04c2cb4adf97897c0f33aace4c40884979bff/src/stream.c#L185-L186
but it's incomplete when a table of strings is passed (it refs the table, but the table values could change and the old strings could then be garbage collected). We need to make sure we ref all the strings that we give to Libuv.

SinisterRectus:
> Is there a use case for allowing numbers here?

Personally, I don't have one. I did find [one case in luvit](https://github.com/luvit/luvit/blob/a7a7eb91ea776c5d02753ad9d03d8c4549e16b50/deps/dns.lua#L682), but these numbers are pre-concatenated. 

>  In my mind passing numbers to one of these write functions is probably a bug, and the coercion was a bug as well. Is your intention keeping backwards compatibility here just in case?

I do agree with this, but I am also thinking about compatibility here. A lot of user-facing luvit and lit functions do not check the data before passing to libuv.

squeek502:
I'm fine with keeping compatibility. The fix will be the same for both strings and numbers coerced to strings, so keeping compat won't be any harder than disallowing numbers.

squeek502:
Superseded by #430

PR #427, opened 2019-10-29
introduce new api luv_optboolean

zhaozg:


src/luv.h
```diff
@@ -142,6 +142,9 @@ static int luv_sock_string_to_num(const char* string);
 static const char* luv_sock_num_to_string(const int num);
 static int luv_sig_string_to_num(const char* string);
 static const char* luv_sig_num_to_string(const int num);
+
+/* from util.c */
+static int luv_optboolean(lua_State*L, int idx, int val);
```
squeek502: Changing the name of `val` to `defaultval` would probably be better.

zhaozg: Ok, wil do

Issue #433, opened 2019-10-31
Maybe unintentional change in how errors are handled


squeek502:
Currently, any runtime errors within a callback don't affect the exit code of Lua (they are just printed to stderr and then execution continues). This seems to have changed in https://github.com/luvit/luv/commit/89f7bafac51706ba8a3280ee06b4225acddee3f9#diff-49ea526d495b77fdd140adde02ced70eL109. Previously, any errors within a callback would lead to `exit(-1)` (the exit was added in https://github.com/luvit/luv/commit/7526c1d9955f085df9eb1c25fcc31c0cfe67e583).

Simple test case:
```lua
local uv = require('luv')

local timer = uv.new_timer()
uv.timer_start(timer, 10, 0, function()
  error('test')
end)

uv.run()
```

Before, running this would cause Lua to exit with status code `-1`, but right now it exits with status code `0`.

This is a major change that might have been unintentional. If it wasn't, then we'll need to figure out the implications (since not exiting on error could lead to weird problems as mentioned in https://github.com/luvit/luv/commit/7526c1d9955f085df9eb1c25fcc31c0cfe67e583).

Related to #127, #128, #192.

@zhaozg

zhaozg:
My original meaning is to avoid the exit of the program in the callback function, and I don't realize the existence of this problem (`uncaught exceptions in event callbacks to avoid undefined behavior in libuv and luajit` ).
So, if possible, is there a way to achieve this, and if not, is it possible to revert to the exit mechanism when the error occurred?

Simply to say, I tend to restore the previous exit to avoid `undefined behavior `.

squeek502:
I think the problem is that there's no realistic way to recover from execution errors in callbacks, since Lua doesn't even have a way to know they happened. I'm away from my computer at the moment so I can't come up with an example, but I will tomorrow.

We can solve this one of two ways:
- Add a method of error handling on the Lua side (this would be something like #127, #128, #192) so that users can decide what to do when a callback has an execution error if they have a way to recover from it
- Exit immediately after an execution error in a callback (this would match how it worked previously)

zhaozg:
@squeek502  A kind and responsible person. :thumbsup:

squeek502:
Here's a somewhat contrived example, but I think it shows the problem.

```lua
local uv = require('luv')

local contents = nil

-- infinite timer that ends when contents is not nil
local timer = uv.new_timer()
timer:start(0, 500, function()
  print("timer tick")
  if contents ~= nil then
[...]

PR #434, opened 2019-11-01
Hard exit on pcall failed

zhaozg:
1. To avoid undefine…d behavior in libuv and luajit follow https://github.com/luvit/luv/commit/7526c1d9955f085df9eb1c25fcc31c0cfe67e583.
2. Replay https://github.com/luvit/luv/issues/433.
3. LUVF_CALLBACK_ flags introduced by https://github.com/luvit/luv/commit/5b68007d6919edd2f5c99830ba0679f958b1235c#diff-4221738e9478598ca8f06be66d888559

### DIFF
--- a/src/work.c
+++ b/src/work.c
@@ -96,7 +96,7 @@ static void luv_work_cb(uv_work_t* req) {
 
   if (lua_isfunction(L, -1)) {
     int i = luv_thread_arg_push(L, &work->arg, 0);
-    i = luv_cfpcall(L, i, LUA_MULTRET, LUVF_CALLBACK_NOEXIT);
+    i = luv_cfpcall(L, i, LUA_MULTRET, 0);
     luv_thread_arg_clear(NULL, &work->arg, 0);
     if ( i>=0 ) {
       //clear in main threads, luv_after_work_cb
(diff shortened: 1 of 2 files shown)


zhaozg:
ci fail with PROCESS_CLEANUP_TEST=1

src/work.c
```diff
@@ -96,7 +96,7 @@ static void luv_work_cb(uv_work_t* req) {
 
   if (lua_isfunction(L, -1)) {
     int i = luv_thread_arg_push(L, &work->arg, 0);
-    i = luv_cfpcall(L, i, LUA_MULTRET, LUVF_CALLBACK_NOEXIT);
+    i = luv_cfpcall(L, i, LUA_MULTRET, 0);
```
squeek502: I think this can still pass `LUVF_CALLBACK_NOEXIT`, since that matches how it worked in 1.9.1:

https://github.com/luvit/luv/blob/976c886cb4ade983dadedc721d9f0071924d5aee/src/work.c#L99-L105

It might lead to the same sort of problems described in https://github.com/luvit/luv/issues/433#issuecomment-548973999 but I'm not familiar enough with the threading side of Luv/Libuv to know.

squeek502:
> ci fail with PROCESS_CLEANUP_TEST=1

This is okay. We can let it fail until I finish #414.

src/work.c
squeek502: On second thought, it's probably better to exit here too. Printing an error and continuing without the user being able to react to the error is probably always a bad idea.

PR #448, opened 2020-01-16
update submodule lua and luajit

zhaozg:


squeek502:
Want to split this into 2 PRs? This one for updating Lua + LuaJIT which we can merge now, and then make a new PR for updating Libuv.

PR #458, opened 2020-02-26
CMake: Add ALWAYS_INSTALL_HEADERS option

### LINKED ISSUE #457: make it possible to install both headers and library
Hi,

I am trying to bump the nixos package for luv and one problem I have is that I can't install both the library and headers at the same time (or I hqven't found how) because BUILD_MODULE is used for 2 distinguished actions
https://github.com/luvit/luv/blob/f38ae04ce02f856566637d2a52527373a5ed61d9/CMakeLists.txt#L131
https://github.com/luvit/luv/blob/f38ae04ce02f856566637d2a52527373a5ed61d9/CMakeLists.txt#L201
On nixos once a package is installed, its folder becomes readonly so that it doesn't get tampered with so I can't rerun the installer twice. I wish there was an other boolean to install the headers (or always install them).

squeek502:
Closes #457

@teto let me know if this is the sort of thing you had in mind.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ set(LUV_VERSION ${LUV_VERSION_MAJOR}.${LUV_VERSION_MINOR}.${LUV_VERSION_PATCH})
 option(BUILD_MODULE "Build as module" ON)
 option(BUILD_SHARED_LIBS "Build shared library" OFF)
 option(WITH_SHARED_LIBUV "Link to a shared libuv library instead of static linking" OFF)
+option(ALWAYS_INSTALL_HEADERS "Install headers regardless of build type" OFF)
 
 if (MINGW)
   add_definitions(-D_WIN32_WINNT=0x0600)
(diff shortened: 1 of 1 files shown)


doronbehar:
I'm the original person who wanted to update our luv on NixOS, (see https://github.com/NixOS/nixpkgs/pull/80528).

Currently, without the patch and with `BUILD_MODULE=OFF`, we get the tree:

```
qvfnhmpb22wyl5zzxynin07lcxn1qy2x-luv-1.34.1-1/
├── include
│   └── luv
│       ├── lhandle.h
│       ├── lreq.h
│       ├── luv.h
│       └── util.h
├── lib
│   ├── libluv.so -> libluv.so.1
│   ├── libluv.so.1 -> libluv.so.1.34.1
│   ├── libluv.so.1.34.1
│   └── pkgconfig
│       └── libluv.pc
└── nix-support
    └── propagated-build-inputs

5 directories, 9 files
```

With the patch and with `BUILD_MODULE=ON` **and** `ALWAYS_INSTALL_HEADERS=ON`, we get the tree:

```
/nix/store/mnv468b533aacm6ji3ywfzlfbg5nr6yc-luv-1.34.1-1
├── include
│   └── luv
│       ├── lhandle.h
│       ├── lreq.h
│       ├── luv.h
│       └── util.h
├── lib
│   └── lua
│       └── 5.1
│           └── luv.so
└── nix-support
    └── propagated-build-inputs

6 directories, 6 files
```

But with _this_ ^ tree, Neovim (e.g) fails to build because it can't find e.g the pkg-config files as found within the 1st tree.

What's so frustrating about the current behavior is that building the Lua module (that ends up in `$prefix/lib/lua/5.1/`) disables the installation of (almost I think) anything else.

Ideally, I'd like to see a fix as follows:

- Make `BUILD_MODULE=ON` install the module **but not disable installation of anything else**!
- Without any new flag introduced, if `BUILD_SHARED_LIBS=ON`, the headers, the pkg-config and whatever else that ended up in `lib/` in the 1st tree will get installed as well.

squeek502:
@doronbehar right now `BUILD_MODULE` and `BUILD_SHARED_LIBS` are expected to be mutually exclusive (if one is on, the other is expected to be off). Are you wanting to be able to build both at once (and install both at once)?

doronbehar:
Correct.

PR #459, opened 2020-02-26
CMake: Make BUILD_MODULE and BUILD_SHARED_LIBS non inclusive

doronbehar:
Running out of patience and in order to get my point better, here's a change I'm satisfied with it's results, here is the resulting `tree` for all 3 configurations of `ON` and `OFF`:

```
prefix-build-module-off-shared-on
└── var
    └── empty
        └── local
            ├── include
            │   └── luv
            │       ├── lhandle.h
            │       ├── lreq.h
            │       ├── luv.h
            │       └── util.h
            └── lib
                ├── libluv.so
                ├── libluv.so.1 -> libluv.so.1.34.1
                ├── libluv.so.1.34.1
                └── pkgconfig
                    └── libluv.pc
prefix-build-module-on-shared-off
└── var
    └── empty
        └── local
            └── lib
                └── lua
                    └── 5.1
                        └── libluv.so
prefix-build-module-on-shared-on
└── var
    └── empty
        └── local
            ├── include
            │   └── luv
            │       ├── lhandle.h
            │       ├── lreq.h
            │       ├── luv.h
            │       └── util.h
            └── lib
                ├── libluv.so
                ├── libluv.so.1 -> libluv.so.1.34.1
                ├── libluv.so.1.34.1
                ├── lua
                │   └── 5.1
                │       └── libluv.so
                └── pkgconfig
                    └── libluv.pc

22 directories, 18 files
```

Fixes https://github.com/luvit/luv/issues/457 and fixes https://github.com/NixOS/nixpkgs/pull/80528 .

NOTE: I've had to create an additional `target` - `libluv` which is distinguished from the `luv` target.


doronbehar:
Wait, that's not what I intended. I'm working on a fix right now.

doronbehar:
Fixed. Updated original comment.

doronbehar:
cc @teto.

CMakeLists.txt
```diff
@@ -157,6 +157,7 @@ if(DEFINED ENV{LUA_COMPAT53_DIR})
 endif()
 include_directories(${LUA_COMPAT53_DIR}/c-api)
 
+if (BUILD_MODULE)
```
doronbehar: Under this condition, exactly the same procedure goes for `libluv` target. Perhaps it's possible to use a `for tmp_target in ["libluv" , "luv"]` instead, but I'm not familiar enough with Cmake yet to write something like that.

squeek502:
Looks good, this is the direction I was thinking of going in as well. Looks like it needs some fixes to get LuaRocks builds working again--I'll work on that.

zhaozg:
Not late, This will break luvi build, we need do more work. @squeek502

squeek502:
> Not late, This will break luvi build, we need do more work. @squeek502

Sorry, you're right. I jumped the gun on this one. `luvi` expects to be able to build a static library by turning `BUILD_MODULE` off, but it's no longer possible to build a static library since [`BUILD_SHARED_LIBS` affects the default library type of `add_library`](https://cmake.org/cmake/help/v3.0/command/add_library.html). This definitely needs some more work.

teto:
another issue we see in https://github.com/NixOS/nixpkgs/pull/80528#discussion_r385013836
 (the `install FILES given no DESTINATION!` part) is that SHAREDLIBS_INSTALL_INC_DIR definition is kinda hard to follow and lacks a good default. Once CMAKE_INSTALL_PREFIX is set can't we set SHAREDLIBS_INSTALL_INC_DIR from that or remove it altogether ?

squeek502:
@teto totally open to suggestions here. I'm not sure I fully understand what the issue is/what the possible fixes are.

doronbehar:
@squeek502 Our build system calls `luarocks make` which in turn calls `cmake` with the following flags:

```
-G"Unix Makefiles"
-DCMAKE_MODULE_LINKER_FLAGS="-shared"
-DLIBDIR="/nix/store/5wf8zrgpp199pmbbgyvkp7hc484pzrbs-luajit-2.1.0-beta3-luv-1.34.2-0/luv-1.34.2-0-rocks/luv/1.34.2-0/lib"
-DLUA_INCDIR="/nix/store/1m8vh1byjnjmjkpdidcqc37cykgsqjap-luajit-2.1.0-beta3/include/luajit-2.1"
-DLUA_LIBDIR=""
-DLUA="/nix/store/1m8vh1byjnjmjkpdidcqc37cykgsqjap-luajit-2.1.0-beta3/bin/luajit"
-DCMAKE_LIBRARY_PATH="/nix/store/1m8vh1byjnjmjkpdidcqc37cykgsqjap-luajit-2.1.0-beta3/lib:/nix/store/w89ndqsb7ig22mw5hk7k7zfvxmhx44xb-libuv-1.34.2/lib:/nix/store/1m8vh1byjnjmjkpdidcqc37cykgsqjap-luajit-2.1.0-beta3/lib:/nix/store/w89ndqsb7ig22mw5hk7k7zfvxmhx44xb-libuv-1.34.2/lib"
-DCMAKE_INCLUDE_PATH="/nix/store/1m8vh1byjnjmjkpdidcqc37cykgsqjap-luajit-2.1.0-beta3/include:/nix/store/w89ndqsb7ig22mw5hk7k7zfvxmhx44xb-libuv-1.34.2/include:/nix/store/1m8vh1byjnjmjkpdidcqc37cykgsqjap-luajit-2.1.0-beta3/include:/nix/store/w89ndqsb7ig22mw5hk7k7zfvxmhx44xb-libuv-1.34.2/include"
-DLUADIR="/nix/store/5wf8zrgpp199pmbbgyvkp7hc484pzrbs-luajit-2.1.0-beta3-luv-1.34.2-0/luv-1.34"
```

Where `/nix/store/5wf8zrgpp199pmbbgyvkp7hc484pzrbs-luajit-2.1.0-beta3-luv-1.34.2-0/` is our `$prefix`.

There are a couple of issues here which can be detected and handled somehow in `CMakeLists.txt` by detecting that `LUA` is set.

1. `LIBDIR` is set to `$prefix/luv-1.34.2-0-rocks/luv/1.34.2-0/lib` which is not where libluv.so should end up, this is where the file loaded by `lua -e "require('lua')"` should end up.
2. The cmake configure phase fails with the following error:

```
CMake Error at CMakeLists.txt:229 (install):
  install FILES given no DESTINATION!
```

I'm not sure my self how to address it yet, but this is my best summarise so I think it's a good start.

doronbehar:
Oh actually, there's even 1 more thing I was thinking of, but consider this one as a nice-to-have:

Since luarocks/luarocks#1160, luarocks/luarocks#509 and luarocks/luarocks#339 are unresolved, perhaps would it be possible to set cmake options via environmental variables? This would help us make the build even cleaner.

teto:
basically we should avoid the scenario 
```
CMake Error at CMakeLists.txt:229 (install):
  install FILES given no DESTINATION!
```
if we follow the definitions, it goes up to `SET(INSTALL_LIB_DIR ${LIBDIR})` but LIBDIR is never defined ?

doronbehar:
> if we follow the definitions, it goes up to SET(INSTALL_LIB_DIR ${LIBDIR}) but LIBDIR is never defined ?

`LIBDIR` _is_ defined, but not to the value we would want it to have.

teto:
where is it defined ? is that some kind of default ? I only can find LUA_LIBDIR.

doronbehar:
It seems like luarocks default, see:

https://github.com/luarocks/luarocks/blob/ec4c5afe2cb70a5ca3e19a5e75d3561316727385/src/luarocks/rockspecs.lua#L63-L80

And:

https://github.com/luarocks/luarocks/blob/ec4c5afe2cb70a5ca3e19a5e75d3561316727385/src/luarocks/path.lua#L108-L119

PR #461, opened 2020-02-29
CMake: Add BUILD_STATIC_LIBS option

squeek502:
@zhaozg this is a followup to https://github.com/luvit/luv/pull/459#issuecomment-592446703

---

This adds on to the changes in https://github.com/luvit/luv/pull/459 to make it possible to build module, static, and shared libraries all at the same time if requested

- To build the module, `-DBUILD_MODULE=On`; the target name for the module is `luv`
- To build a static library, `-DBUILD_STATIC_LIBS=On`; the target name for the static lib is `libluv_a` (this matches how `libuv` names its targets [`uv` for shared and `uv_a` for static])
- To build a shared library, `-DBUILD_SHARED_LIBS=On`; the target name for the shared lib is `libluv`

These are able to be mixed/matched as needed.

This is a breaking change (as was #459) for users of CMake (e.g. luvi) who will now need to specify linking against `libluv`/`libluv_a` and turn on `BUILD_SHARED_LIBS`/`BUILD_STATIC_LIBS` before calling `add_subdirectory`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ set(LUV_VERSION_PATCH 2)
 set(LUV_VERSION ${LUV_VERSION_MAJOR}.${LUV_VERSION_MINOR}.${LUV_VERSION_PATCH})
 
 option(BUILD_MODULE "Build as module" ON)
+option(BUILD_STATIC_LIBS "Build static library" OFF)
 option(BUILD_SHARED_LIBS "Build shared library" OFF)
 option(WITH_SHARED_LIBUV "Link to a shared libuv library instead of static linking" OFF)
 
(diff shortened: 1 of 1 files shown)


squeek502:
Some notes:

- It might be possible to fix this being a breaking change when exactly one build target is specified, using `luv` as the target name for whatever library type is specified.
  + For example, luvi sets `BUILD_MODULE=Off` and then expects `luv` to be a static library. This could be maintained by setting the static library's target name `luv` when both `BUILD_MODULE=Off` and `BUILD_SHARED_LIBS=Off` (and also infer `BUILD_STATIC_LIBS=On` from the other two being off)
- If anyone has other ideas for how to allow multiple library types to be built at the same time in a less breaking way, let me know

PR #464, opened 2020-03-14
CMake: Fix header install paths for shared/static builds when using LuaRocks

### LINKED ISSUE #463: SHAREDLIBS_INSTALL_INC_DIR not set
Still trying to package libuv on nixos
```
-- Lua: using information from luarocks
-- LUA_LIBDIR: 
-- LUA_INCDIR: /nix/store/491m3i5n0l92aar68hjy0zv61yw11jc6-luajit-2.1.0-beta3/include/luajit-2.1
-- LUA: /nix/store/491m3i5n0l92aar68hjy0zv61yw11jc6-luajit-2.1.0-beta3/bin/luajit
-- Lua library: LUA_LIBRARIES-NOTFOUND
-- path 
CMake Error at CMakeLists.txt:253 (install):
  install FILES given no DESTINATION!
```
Going through the various steps, `SHAREDLIBS_INSTALL_INC_DIR` ,  I find `set(SHAREDLIBS_INSTALL_INC_DIR "${INSTALL_INC_DIR}")` but INSTALL_INC_DIR is possibly never set (hence the error), whereas   INSTALL_LIB_DIR defaults to ${LIBDIR}. Would be cool to have a nice default.

squeek502:
Fixes #463

@teto let me know if this fixes the problem for you.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,23 +202,27 @@ if (NOT LUA)
   if (BUILD_STATIC_LIBS)
     set(STATICLIBS_INSTALL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib"
       CACHE PATH "Installation directory for static libraries")
-    set(STATICLIBS_INSTALL_INC_DIR "${CMAKE_INSTALL_PREFIX}/include/luv"
-      CACHE PATH "Installation directory for headers")
   endif (BUILD_STATIC_LIBS)
   if (BUILD_SHARED_LIBS)
     set(SHAREDLIBS_INSTALL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib"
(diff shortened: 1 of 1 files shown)


teto:
this seems to have fixed it. I wonder if you could add a `rockspec_format = "3.0"` to the rockspec ? 
I think I will probably need to patch the rockspec to work around nix limitations anyway so no worry if you dont want to.
NB: I intend to work around limitations in nix/luarocks by disabling luv build and install pass.
```
build = {
  type = 'cmake',
  build_pass = false,
  install_pass = false,
```

squeek502:
I think I'd rather not specify `rockspec_format = "3.0"` unless we need to (we are still testing against LuaRocks 2.3.0 in CI).

teto:
fantastic thanks !

PR #479, opened 2020-04-16
Update to libuv 1.36.0

squeek502:
- https://github.com/libuv/libuv/releases/tag/v1.36.0
- http://docs.libuv.org/en/v1.x/fs.html#c.uv_fs_lutime


squeek502:
We seem to be getting random failures in CI when downloading the LuaRocks 2.3 sources:

```
+curl --silent --location http://luarocks.org/releases/luarocks-2.3.0.tar.gz
+tar xz
gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

PR #480, opened 2020-04-16
CI: setup_lua: use https for source tar downloads

squeek502:
I'm kind of hoping this might fix the random CI failures when downloading LuaRocks source tars mentioned [here](https://github.com/luvit/luv/pull/479#issuecomment-614509413), but using `https` urls makes sense anyway.


zhaozg:
sometimes, I got ERR_CONNECT_TIMEOUT when access https://luarocks.org/ too

PR #487, opened 2020-04-22
luv_new_udp: Add support for UV_UDP_RECVMMSG

squeek502:
See the added comment for my thought process here:

https://github.com/luvit/luv/blob/43fb1183d4599e06ad85df1fc8a4b2782f35a0c4/src/udp.c#L48-L57

Submitting this as a draft because there's still some more things to be figured out:

- There are some bugs that need to be worked out with UV_UDP_RECVMMSG enabled, some of the udp tests are failing
- I think the usefulness of `recvmmsg` depends on the buffer size, which we don't give Lua control of (we use the `suggested_size` from [uv_alloc_cb](http://docs.libuv.org/en/v1.x/handle.html#c.uv_alloc_cb)). Maybe turning on `UV_UDP_RECVMMSG` only makes sense if we also allow changing the size of the recv buffer.
- The parameter of `luv_new_udp` is a little bit strange now, I think we might want to allow it to be passed as a table and have different fields for address family and recvmmsg. This comment is relevant: https://github.com/luvit/luv/issues/399#issuecomment-540185444

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -1888,12 +1888,26 @@ UDP handles encapsulate UDP communication for both clients and servers.
 ### `uv.new_udp([flags])`
 
 **Parameters:**
-- `flags`: `string` or `nil`
+- `flags`: `table` or `nil`
+  - `family`: `string` or `nil`
+  - `mmsgs`: `integer` or `nil` (default: `1`)
 
 Creates and initializes a new `uv_udp_t`. Returns the Lua userdata wrapping
-it. The actual socket is created lazily. Flags may be a family string:
-`"unix"`, `"inet"`, `"inet6"`, `"ipx"`, `"netlink"`, `"x25"`, `"ax25"`,
-`"atmpvc"`, `"appletalk"`, or `"packet"`.
+it. The actual socket is created lazily.
+
+When specified, `family` must be one of `"unix"`, `"inet"`, `"inet6"`,
+`"ipx"`, `"netlink"`, `"x25"`, `"ax25"`, `"atmpvc"`, `"appletalk"`, or
+`"packet"`.
+
+When specified, `mmsgs` determines the number of messages able to be received
+at one time via `recvmmsg(2)` (the allocated buffer will be sized to be able
+to fit the specified number of max size dgrams). Only has an effect on
+platforms that support `recvmmsg(2)`.
+
+**Note:** For backwards compatibility reasons, `flags` can also be a string or
+integer. When it is a string, it will be treated like the `family` key above.
+When it is an integer, it will be used directly as the `flags` parameter when
+calling `uv_udp_init_ex`.
 
 **Returns:** `uv_udp_t userdata` or `fail`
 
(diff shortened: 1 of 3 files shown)


squeek502:
Some part of the test failures are from libuv bugs with `UV_UDP_RECVMMSG`. Upstream issue: https://github.com/libuv/libuv/issues/2806

squeek502:
Valgrind/ASAN test failures might also be a libuv upstream issue: https://github.com/libuv/libuv/pull/2818

squeek502:
More relevant info regarding

> I think the usefulness of recvmmsg depends on the buffer size, which we don't give Lua control of (we use the suggested_size from uv_alloc_cb). Maybe turning on UV_UDP_RECVMMSG only makes sense if we also allow changing the size of the recv buffer.

[from my comment here]():

> * `UV_UDP_RECVMMSG` has no real effect unless the allocated buffer size is large enough to handle > 1 max size dgrams
> * When `UV_UDP_RECVMMSG` is used, the number of possible messages recv'd at once is limited to (buffer size / max dgram size).

`suggested_size` is equal to max dgram size, so without altering the allocated buffer size (or giving control of the buffer size to Lua), turning on `UV_UDP_RECVMMSG` in luv will have no effect.

squeek502:
More relevant info regarding

> I think the usefulness of recvmmsg depends on the buffer size, which we don't give Lua control of (we use the suggested_size from uv_alloc_cb). Maybe turning on UV_UDP_RECVMMSG only makes sense if we also allow changing the size of the recv buffer.

[from my comment here](https://github.com/libuv/libuv/pull/2818#issuecomment-620846742):

> * `UV_UDP_RECVMMSG` has no real effect unless the allocated buffer size is large enough to handle > 1 max size dgrams
> * When `UV_UDP_RECVMMSG` is used, the number of possible messages recv'd at once is limited to (buffer size / max dgram size).

`suggested_size` is equal to max dgram size, so without altering the allocated buffer size (or giving control of the buffer size to Lua), turning on `UV_UDP_RECVMMSG` in luv will have no effect.

squeek502:
Ok, this is ready for review. I decided against enabling `recvmmsg` by default and instead the user has to opt in by setting the number of messages that can be received at a time by calling e.g. `uv.new_udp({mmsgs = 4})` (to be able to receive 4 messages at once). See the comment added to where the default for `mmsgs` is set:

https://github.com/luvit/luv/blob/407f9c380bd2dc3db6766834865962d05c9f8723/src/udp.c#L31-L37

(in some very basic testing, I noticed that using recvmmsg has some overhead so I figured always enabling it is probably not the correct decision; I don't know much about its use case, though, so if anyone has more experience let me know what you think)

All the rest is handled automatically, though, and there's no difference from the Lua side (the same number of Lua recv callbacks are called for a udp handle that is using recvmmsg as one that isn't).

zhaozg (approved):

Issue #410, opened 2019-10-13
Remaining unbound functions


squeek502:
Not an exhaustive list yet, still a WIP:

- [ ] [`uv_stream_get_write_queue_size`](http://docs.libuv.org/en/v1.x/stream.html#c.uv_stream_get_write_queue_size); see also the non-standard `luv` function `tcp_write_queue_size`, presumably bound before `uv_stream_get_write_queue_size` was added
- [ ] [`uv_loop_configure`](http://docs.libuv.org/en/v1.x/loop.html#c.uv_loop_configure) (#103)
- [ ] [`uv_setup_args`](http://docs.libuv.org/en/v1.x/misc.html#c.uv_setup_args) (#264); is this bindable?

SinisterRectus:
We could probably deprecate tcp_write_queue_size and getpid if they have libuv versions.

lchown needs implementing.

squeek502:
What should be the process for deprecation?

SinisterRectus:
I would just leave a comment in the code and document that it should not be used, or just not document it at all. We can't remove anything without a major version bump.

squeek502:
Added to the OP: an autogenerated a list of all functions defined in `uv.h` that match the pattern `UV_EXTERN return_type uv_function_name()` that have no occurrences in any of Luv's C files. It's bigger than I was expecting, and it's likely that we don't really want to bind them all. Many have to do with threading/synchronization.

zhaozg:
Brave man

nikneym:
Are there any plans to exposing loop utilities to Lua? I'm in a situation where I need to use `fork` syscall but in order to do that I have to call `uv_loop_fork`, where I need a pointer to loop. I know this is a niche case but it might still be effective to support loop calls.

PR #491, opened 2020-05-11
udp: Add udp_set_source_membership binding + improve udp_set_membership

squeek502:
- Allows the interface argument of udp_set_membership to be nil (meaning a NULL const char* passed to uv_udp_set_membership). This was needed to be able to port the libuv multicast join tests
- Adds a binding for udp_set_source_membership (contributes towards #410)
- Ports the test-udp-multicast-join and test-udp-multicast-join6 tests from libuv

---

Note that the added test cases might not be portable. Libuv has some special handling for specific platforms that hasn't been ported to luv:

- https://github.com/libuv/libuv/blob/55dede5b6f730882d79f994138f0f23a7b28b119/test/test-udp-multicast-join.c#L129-L132
- https://github.com/libuv/libuv/blob/55dede5b6f730882d79f994138f0f23a7b28b119/test/test-udp-multicast-join6.c#L33-L44

EDIT: These might be solved by https://github.com/luvit/luv/pull/491/commits/46572c3ff8bc40372648f7252c0c4bc6922be86c and https://github.com/luvit/luv/pull/491/commits/356968cdf7f8cdd53ce932c9d7f74043adb7ee86

EDIT#2: Forced the OSX jobs to run and confirmed that the tests pass there: https://travis-ci.org/github/luvit/luv/builds/685548492

### DIFF
--- a/.ci/bindcov.sh
+++ b/.ci/bindcov.sh
@@ -26,7 +26,7 @@ skipped+=(
 	uv_loop_size uv_loop_fork uv_loop_get_data uv_loop_set_data uv_strerror uv_strerror_r uv_err_name
 	uv_err_name_r uv_handle_size uv_handle_get_type uv_handle_type_name uv_handle_get_data
 	uv_handle_get_loop uv_handle_set_data uv_req_size uv_req_get_data uv_req_set_data uv_req_get_type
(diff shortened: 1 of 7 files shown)


squeek502:
Valgrind failure is a libuv bug (same failure happens when running the Libuv test directly), will file an upstream issue.

Issue #497, opened 2020-05-12
Lua 5.4


SinisterRectus:
Lua 5.4.0 is nearing a final release. It is currently at [rc3](http://lua-users.org/lists/lua-l/2020-05/msg00043.html).

Our supported Lua versions according to our rockspec is `lua >= 5.1`. Does this imply that 5.4 is supported on release? If so, think we should should look at whether that is true. I'm not sure if there are any compatibility changes that will affect us.

squeek502:
Would definitely be nice to start testing 5.4 in CI.

SinisterRectus:
Lua 5.4 has formally released since opening this issue. https://www.lua.org/manual/5.4/readme.html

SinisterRectus:
Looks like #515 is merged. Are the build issues fixed? Should we make a 1.14.0 release?

zhaozg:
> Looks like #515 is merged. Are the build issues fixed? Should we make a 1.40.0 release?

YES

squeek502:
I'd like to finish up #487 and merge it before releasing 1.40.0. Will try to do that soon.

zhaozg:
Anyone need to do before a new release

SinisterRectus:
There are some open issues. #503 / #505 / #506. Not sure how easy those are to fix.

PR #504, opened 2020-07-02
Update Lua version check

javalikescript:
Building using the new Lua 5.4 version fails due to the included file `compat-5.3.h`

### DIFF
--- a/src/luv.c
+++ b/src/luv.c
@@ -16,7 +16,7 @@
  */
 
 #include <lua.h>
-#if (LUA_VERSION_NUM != 503)
+#if (LUA_VERSION_NUM < 503)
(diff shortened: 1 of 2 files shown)


SinisterRectus:
Looks okay to me, but I think we need to make a formal effort to support 5.4. See #497.

javalikescript:
I would love _luv_ to fully support _Lua_ 5.4! The goal of this pull request is only to allow compilation.
I use a subset of _luv_ with the latest _Lua_ version and I have my own tests. This change allow me to continue to reference the _luvit/luv_ repository waiting for a more broader support.

zhaozg:
And should update submodule lua compat 5.3

PR #509, opened 2020-08-31
Update submodule(LuaJIT,libuv) version

zhaozg:


zhaozg:
unbind api
* uv_library_shutdown
* uv_udp_using_recvmmsg
* uv_metrics_idle_time
* uv_fs_get_system_error

squeek502:
`uv_library_shutdown` should be added to the skipped variable here:

https://github.com/luvit/luv/blob/2eee758b2affb76dc623ca6119a334008a4fc4be/.ci/bindcov.sh#L9-L10

`uv_udp_using_recvmmsg` should be skipped for now ([added to this list](https://github.com/luvit/luv/blob/2eee758b2affb76dc623ca6119a334008a4fc4be/.ci/bindcov.sh#L22-L33)). It will be useful for finishing https://github.com/luvit/luv/pull/487 but unfortunately https://github.com/libuv/libuv/pull/2836 didn't get included in 1.39.0 and I'd like to wait for that to be merged before dealing with recvmmsg in luv.

`uv_metrics_idle_time` should be bound. Note that the [`uv_loop_configure`](http://docs.libuv.org/en/v1.x/loop.html#c.uv_loop_configure) binding also needs to be updated.

`uv_fs_get_system_error` could be bound, since `uv_fs_t` userdata is returned from our fs async functions.

zhaozg:
Ok, let me try.

squeek502:
Made a PR to your branch to add some of the bindings: https://github.com/zhaozg/luv/pull/2

squeek502:
I think it's okay to skip binding `uv_fs_get_system_error` actually, because there's not really any easy way to use it from Lua that I can find with how we have things set up.

For example, in the following test code:

```lua
    local req
    req = uv.fs_open('some-file-that-doesnt-exist', 'r', tonumber('644', 8), function(err, fd)
      p("in cb", req)
      p(uv.fs_get_system_error(req))
    end)
    p("before cb", req)
    p(uv.fs_get_system_error(req))
```
It's inconsistent whether or not `fs_get_system_error` will give 0 or an error code before the callback is called (meaning it's not all that useful), and then once the callback is called, the uv_fs_t userdata no longer gets identified as a uv_fs_t and throws the following error:

```
Uncaught Error: tests/test-fs.lua:339: bad argument #1 to 'fs_get_system_error' (Expected uv_fs_t)
stack traceback:
	[C]: in function 'luv.fs_get_system_error'
	tests/test-fs.lua:339: in function <tests/test-fs.lua:337>
	[C]: in function 'luv.run'
	./lib/tap.lua:61: in function <./lib/tap.lua:48>
	[C]: in function 'xpcall'
	./lib/tap.lua:48: in upvalue 'run'
	./lib/tap.lua:146: in function 'lib/tap'
	(...tail calls...)
	[C]: in ?
```

This error might be fixable, depending on what exactly is causing it, but `luv` in general hides `uv_req_t` stuff from the Lua side of things, so I'm not sure something like `fs_get_system_error` is worth binding for now. Plus, uv_fs_t userdata is only ever returned from the async versions of fs functions; if the sync version is used instead, it's not possible to access the `uv_fs_t` req, meaning `fs_get_system_error` is never callable.

However, it should probably be added to #410 because things like `uv_fs_get_type` and `uv_fs_get_path` would likely have similar problems, so addressing them all at once would make sense (either intentionally leaving them all unbound or making binding them make more sense).

squeek502 (approved):

Issue #510, opened 2020-09-01
Memory bugs from garbage collection


squeek502:
From https://github.com/luvit/luv/issues/507#issuecomment-685126443:

> Found a consistent minimal test case by making the garbage collector extremely aggressive.
> 
> In fact, tuning the garbage collector to be this aggressive also makes our existing `tests/test-thread.lua` reproduce this bug as well (i.e. `valgrind lua -e "collectgarbage('setpause', 0); collectgarbage('setstepmul', 10000000000000)" tests/test-thread.lua`).
> 
> I'm thinking we should either add a new CI step that runs the tests through valgrind with extremely aggressive GC or alter our existing valgrind/ASAN steps to use extremely aggressive GC. It will likely surface some more bugs.

I just tried running our full test suite through `valgrind lua -e "collectgarbage('setpause', 0); collectgarbage('setstepmul', 10000000000000)" tests/run.lua`. Here are the bugs that were found:

---

- [x] process - get pid, process - spawn and kill by handle, process - spawn and kill by pid (all have the same error): https://github.com/luvit/luv/pull/511

```
==8329== Syscall param execve(argv[i]) points to unaddressable byte(s)
==8329==    at 0x4AA116B: execve (syscall-template.S:78)
==8329==    by 0x4AA1B5E: __execvpe_common (execvpe.c:136)
==8329==    by 0x4AA1B5E: execvpe (execvpe.c:190)
==8329==    by 0x4FE0BEE: uv__process_child_init (process.c:401)
==8329==    by 0x4FE0BEE: uv_spawn (process.c:500)
==8329==    by 0x4FD1080: luv_spawn (process.c:268)
==8329==    by 0x114CC5: luaD_precall (ldo.c:434)
==8329==    by 0x118E60: luaV_execute (lvm.c:1134)
==8329==    by 0x114E37: luaD_call (ldo.c:499)
==8329==    by 0x114E9A: luaD_callnoyield (ldo.c:509)
==8329==    by 0x114E9A: f_call (lapi.c:943)
==8329==    by 0x110492: luaD_rawrunprotected (ldo.c:142)
==8329==    by 0x113E1A: luaD_pcall (ldo.c:729)
==8329==    by 0x12406D: lua_pcallk (lapi.c:969)
==8329==    by 0x129C08: luaB_xpcall (lbaselib.c:441)
==8329==  Address 0x70c9008 is 24 bytes inside a block of size 26 free'd
[...]

PR #511, opened 2020-09-01
Fix args potentially being garbage collected before use in luv_spawn

squeek502:
This is one possible fix for the process bug in https://github.com/luvit/luv/issues/510. One other way to fix it would be to just leave the args strings on the stack instead of popping them off. Not sure which is better.


zhaozg (approved):

Timothy-:
I have the following error due to this commit:

```
src/process.c:51: error: 'for' loop initial declarations are only allowed in C99 mode
```

squeek502:
Whoops, I always forget about that. Will fix.

PR #520, opened 2020-10-20
Add BUILD_STATIC_LIBS to Makefile

squeek502:
Should have been added in https://github.com/luvit/luv/pull/461/ when BUILD_STATIC_LIBS option was added


rphillips (approved):

mnemnion:
I'd suggest updating the README to reflect the new way of building static libs as well.

PR #528, opened 2021-02-11
fs: Print destination paths in error messages for the relevant functions

squeek502:
Output of the added test, before (note that the relevant path for each of these errors messages is actually the destination, not the source):
```
  "fs_rename"	"EISDIR: illegal operation on a directory: luv7TbDmb"
  "fs_link"	"EEXIST: file already exists: luv7TbDmb"
  "fs_symlink"	"EEXIST: file already exists: luv7TbDmb"
  "fs_copyfile"	"EISDIR: illegal operation on a directory: luv7TbDmb"
```
After:
```
  "fs_rename"	"EISDIR: illegal operation on a directory: luv7TbDmb -> luvQzLgbb"
  "fs_link"	"EEXIST: file already exists: luv7TbDmb -> luvQzLgbb"
  "fs_symlink"	"EEXIST: file already exists: luv7TbDmb -> luvQzLgbb"
  "fs_copyfile"	"EISDIR: illegal operation on a directory: luv7TbDmb -> luvQzLgbb"
```

### DIFF
--- a/src/fs.c
+++ b/src/fs.c
@@ -208,6 +208,20 @@ static void luv_push_statfs_table(lua_State* L, const uv_statfs_t* s) {
 };
 #endif
 
+static int fs_req_has_dest_path(uv_fs_t* req) {
+  switch (req->fs_type) {
+    case UV_FS_RENAME:
+    case UV_FS_SYMLINK:
+    case UV_FS_LINK:
+#if LUV_UV_VERSION_GEQ(1, 14, 0)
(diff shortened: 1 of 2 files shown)


creationix (changes requested):
Looks good, but I think you're only handling the sync case.  The errors for async callbacks still have the old behavior.  Look online 224.

squeek502:
Good catch, will fix that.

creationix (approved):

PR #532, opened 2021-02-19
Make symlink/copyfile bindings more user-friendly

### LINKED ISSUE #529: fs_symlink and fs_copyfile don't fail when flags param is wrong
Related to https://github.com/luvit/lit/pull/280, this type of error is harder to identify because currently the symlink/copyfile functions accept a function in the flags argument without failing. I think there are two options here:

1. Error when the flags argument is an unexpected type
2. Treat the 3rd argument as the callback if we detect that the flags argument has been omitted

I haven't double-checked, but I feel like there's precedent for both of these options in our current bindings. Not sure which would be better, but I think either of them is better than the current silent failure.

squeek502:
- Callback for the async version can be given as the 3rd parameter if the flags argument is omitted
- Flags can be specified as either a table, an integer, or nil, and will throw an error for unexpected types

Closes #529

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -2762,19 +2762,19 @@ Equivalent to `link(2)`.
 
 **Returns (async version):** `uv_fs_t userdata`
 
-### `uv.fs_symlink(path, new_path, flags, [callback])`
+### `uv.fs_symlink(path, new_path, [flags], [callback])`
 
 **Parameters:**
 - `path`: `string`
 - `new_path`: `string`
-- `flags`: `table` or `nil`
+- `flags`: `table`, `integer`, or `nil`
   - `dir`: `boolean`
   - `junction`: `boolean`
 - `callback`: `callable` (async version) or `nil` (sync version)
   - `err`: `nil` or `string`
   - `success`: `boolean` or `nil`
 
-Equivalent to `symlink(2)`.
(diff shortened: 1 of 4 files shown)


squeek502:
Example of how they can now be called and the possible errors:

```lua
local uv = require('luv')

-- sync
uv.fs_symlink('from', 'to')
uv.fs_symlink('from', 'to', {})
uv.fs_symlink('from', 'to', {}, nil)

-- async
uv.fs_symlink('from', 'to', function() end)
uv.fs_symlink('from', 'to', 0, function() end)
uv.fs_symlink('from', 'to', {}, function() end)
uv.fs_symlink('from', 'to', nil, function() end)

-- bad argument errors
print(pcall(function() uv.fs_symlink('from', 'to', function() end, function() end) end))
print(pcall(function() uv.fs_symlink('from', 'to', function() end, nil) end))
print(pcall(function() uv.fs_symlink('from', 'to', nil, {}) end))

uv.run()
```

Outputs:

```
false	tmp.lua:15: bad argument #3 to 'fs_symlink' (expected nil, integer, or table; got function)
false	tmp.lua:16: bad argument #3 to 'fs_symlink' (expected nil, integer, or table; got function)
false	tmp.lua:17: bad argument #4 to 'fs_symlink' (function or callable table expected, got table)
```

SinisterRectus (approved):
LGTM, though I haven't tested the code.

PR #534, opened 2021-04-12
Update libuv to 1.41.0

creationix:


squeek502:
New functions:
- [`uv_pipe`](http://docs.libuv.org/en/v1.x/pipe.html#c.uv_pipe)
- [`uv_socketpair`](http://docs.libuv.org/en/v1.x/tcp.html?highlight=uv_socketpair#c.uv_socketpair)

(for some reason `.ci/bindcov.sh` doesn't report uv_pipe as a missing binding, will try to look into that)

zhaozg (commented):
Good PR.

zhaozg:
Are you plan to do uv_socketpair?

squeek502:
Yes, but I haven't decided how to bind it yet, since the `type` and `protocol` parameters are slightly weird. Here's what I'm thinking currently:

`type`:
- This could potentially be a string, where the options are `stream`, `dgram`, `seqpacket`, `raw`, `rdm`
- [Windows docs (see `type` section)](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsasocketw)
- [Linux/Unix docs](https://man7.org/linux/man-pages/man2/socket.2.html)
- The Libuv test suite only uses `SOCK_STREAM` for this parameter

`protocol`:
- This one seems tough to bind in a user-friendly way, since it relies on `type` and `domain`, so we might just need to leave it as an integer
- The Libuv test suite only uses `0` for this parameter
- Note: on Windows, `AF_INET` is used as the domain when creating the sockets; on Unix, `AF_UNIX` is used as the domain

So maybe something like this?

```markdown
uv.socketpair([type], [protocol], [flags0], [flags1])

**Parameters:**
- `type`: `string` or `nil` (default: `stream`)
- `protocol`: `integer` or `nil` (default: 0)
- `flags0`: `table` or `nil`
  - `nonblock`: `boolean` (default: `false`)
- `flags1`: `table` or `nil`
  - `nonblock`: `boolean` (default: `false`)
```

Let me know if you have any better ideas.

zhaozg:
Look https://man7.org/linux/man-pages/man2/socketpair.2.html, and https://man7.org/linux/man-pages/man2/socket.2.html, I prefer to reuse the code for `type` and `protocol` in https://github.com/luvit/luv/blob/b3a7e095d274ed6e467603f2ae594295f25c7574/src/dns.c#L118-L148
for socket() we should keep open accpet wide value range, If not right, libuv will report error in runtime.

squeek502:
I think we're actually binding the `protocol` incorrectly in `getaddrinfo`. It treats it like an `AF_` constant, but those are separate from the protocol constants.

From [`socket(2)`](https://man7.org/linux/man-pages/man2/socket.2.html):

> The protocol specifies a particular protocol to be used with the socket. Normally only a single protocol exists to support a particular socket type within a given protocol family, in which case protocol can be specified as 0. However, it is possible that many protocols may exist, in which case a particular protocol must be specified in this manner. The protocol number to use is specific to the “communication domain” in which communication is to take place; see protocols(5). See getprotoent(3) on how to map protocol name strings to protocol numbers.

From [`protocols(5)`](https://man7.org/linux/man-pages/man5/protocols.5.html):

> This file is a plain ASCII file, describing the various DARPA internet protocols that are available from the TCP/IP subsystem.

From my local `/etc/protocols` file:

> `udp	17	UDP		# user datagram protocol`

And from the output of `test-dns.lua`

> `{ addr = "206.189.225.172", protocol = "packet", family = "inet", socktype = "dgram" }`

That is, luv is translating the returned protocol `17` into `"packet"` (because `AF_PACKET` is 17) here:

https://github.com/luvit/luv/blob/b3a7e095d274ed6e467603f2ae594295f25c7574/src/dns.c#L52

but that is incorrect; it is actually the `udp` protocol. If we still want to translate between string <-> number, we should probably be using [`getprotobyname`/`getprotobynumber`](https://man7.org/linux/man-pages/man3/getprotoent.3.html) instead, but I'm not sure if there's something similar on Windows.

I'll try implementing how I think protocols should be handled when binding `socketpair`, but I'll leave `getaddrinfo` alone and submit a PR afterwards to fix how that handles protocols.

zhaozg (approved):

creationix:
Nice work guys!  Also I'm not surprised I got the protocol stuff wrong.  I barely understood it at the time.

PR #540, opened 2021-04-20
Fix udp multicast test when no exernal interfaces are available

### LINKED ISSUE #539: Failing lua luv tests in build environments
Hi,

test in tests/test-udp.lua seems to fail in build environments with EPERM with limited network. The fedora build hosts don't have internet access, so don't have any network configuration. Thus luv tests are failing.

See e.g.
https://kojipkgs.fedoraproject.org//work/tasks/949/66240949/build.log

I would comment them out for now.

Maybe https://cwrap.org/socket_wrapper.html would help to improve network testing :-)

squeek502:
Fixes #539


zhaozg (approved):

squeek502:
Whoops, should have closed this since this doesn't actually fix #539; that'll need a different solution. The changes here don't really hurt anything though.

PR #551, opened 2021-06-30
Add Github Actions workflow

squeek502:
~~Still WIP, just thought I'd put up a draft PR for now~~

EDIT: All of this comment is now outdated, see comment(s) below instead

- Will clean up the commit history once things are more finalized
- Valgrind is reporting a leak with threads (test-thread.lua and test-async.lua trigger it), but I don't understand what's happening--I can't tell if its real or a false positive. Will post more details on this as I investigate it more. The full stacktrace (for some reason the symbols aren't being loaded properly in CI) is:
```
==10602== 272 bytes in 1 blocks are possibly lost in loss record 4 of 6
==10602==    at 0x483DD99: calloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==10602==    by 0x40149CA: allocate_dtv (dl-tls.c:286)
==10602==    by 0x40149CA: _dl_allocate_tls (dl-tls.c:532)
==10602==    by 0x5025322: allocate_stack (allocatestack.c:622)
==10602==    by 0x5025322: pthread_create@@GLIBC_2.2.5 (pthread_create.c:660)
==10602==    by 0x4FFB46D: uv_thread_create_ex (thread.c:259)
==10602==    by 0x4FD8BED: luv_new_thread (thread.c:325)
==10602==    by 0x12C1CA: luaD_call (ldo.c:482)
==10602==    by 0x1370AC: luaV_execute (lvm.c:1615)
==10602==    by 0x12C371: luaD_call (ldo.c:504)
==10602==    by 0x1370AC: luaV_execute (lvm.c:1615)
==10602==    by 0x12C371: luaD_call (ldo.c:504)
==10602==    by 0x12C46E: luaD_callnoyield (ldo.c:526)
==10602==    by 0x13B08B: f_call (lapi.c:997)
```
- No deploy phase yet, will work on that after basic CI is in place
- Looks like I'm not doing the matrix right, it's only doing one combo of lua version/luarocks version instead of 4

Will close #550 once this is done.

PR #556, opened 2021-07-03
Add deploy step to Github Actions

squeek502:
Follow up to #551

- Only run on tagged commits
- Should behave the same as it did on Travis
- Needs LUAROCKS_API_KEY set as a secret for the repository: https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository


zhaozg (approved):

creationix:
Secret added.

PR #557, opened 2021-07-13
Add flag to generate compile_commands.json

otherJL0:
Adding an option to CMakeLists.txt to generate compilation database (compile_commands.json) during compilation for when working with clangd


squeek502:
I have two potential hangups:
- Is this necessary to have committed to the repo rather than just having those that use clangd handle it locally? Looks like `CMAKE_EXPORT_COMPILE_COMMANDS` can be set via an environment variable, so it being in the CMakeLists.txt is not totally required
- Having `CMAKE_EXPORT_COMPILE_COMMANDS` in the CMakeLists.txt probably bumps the minimum CMake version to 3.5: https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html

otherJL0:
While clangd users can manually configure their local environment, I would like to explicitly make the project clangd friendly to reduce the development friction as much as possible. 

I overlooked the minimum required CMake version and unfortunately I do think the version would need to be bumped to 3.5 if adding the flag. Bumping the version from 2.8 to 3.5 seems like a big leap and I wouldn't want to disrupt others' workflows.

I'm still new to CMake, but I'll try and modify the file to only set the `CMAKE_EXPORT_COMPILE_COMMANDS` flag if the detected version is at least 3.5 as a safer workaround if deemed the more appropriate route.

One other hangup I've noticed is that once clangd is enabled, it tries to format the c files using its default settings. To avoid definitively setting a formatting style for the project, I've also added a minimal `.clang-format` file to disable formatting. Even with formatting disabled, the imports are still sorted which should be less controversial and are included along with the `.clang-format` commit.

otherJL0:
Added version check to CMakeLists.txt to determine if version is high enough to export the project's compile commands database

squeek502 (approved):
I'm cool with this if others are.

PR #560, opened 2021-07-25
Add docs coverage script + fix documentation using what it found

squeek502:
Adds a docscov.sh script that runs in CI to check `luv.c` against `docs.md` to see if there are any functions that are defined in luv.c but not documented in docs.md. Compliments the existing bindcov.sh and will help make sure we document any newly bound functions in the future.

Also updates `docs.md` to fix all the undocumented functions that docscov.sh initially found.


SinisterRectus (approved):
Cool.

PR #562, opened 2021-08-08
fix #561, uninitialised memory access

### LINKED ISSUE #561: Thread arg does not support type
```lua
function M.test_threading()
  local work_ctx = vim.loop.new_work(function(...)
    local args = {...}
  end, function(arg)
  end)

  work_ctx:queue()
end

function M.init()
  for i=0, 100 do
    M.test_threading()
  end
end
```
So I am using this code in neovim, and every time I run init, I get a new error:
`15780 Segmentation fault`
`Error: thread arg not support type HG H9G, 15865 Segmentation fault`
`Error: thread arg not support type (null), 16475 Segmentation fault`
`Error: thread arg not support type no value, 16774 Segmentation fault`
`Error: thread arg not support type ATUHHHGD$, 17714 Segmentation fault`
`Error: thread arg not support type UHw(HHG H9, 18485 Segmentation fault`
...

zhaozg:


squeek502:
This will fix the segfaults, but is it actually okay for `luv_cfpcall` to be returning `-1` when there are 0 values returned by the work_cb function?

https://github.com/luvit/luv/blob/727c6e45f7f6414a95cf7e63443d7aa42a4c9481/src/work.c#L96

(more details at https://github.com/luvit/luv/issues/561#issuecomment-894635669)

zhaozg:
> This will fix the uninitialized reads, but is it actually okay for `luv_cfpcall` to be returning `-1` when there are 0 values returned by the work_cb function?

that because vm stack index(get_top after push err_funcs) not balance, fixed in 
https://github.com/luvit/luv/pull/562/commits/2d685c3a2c3abd6e4da0dd9dedc1e22ffe856851

thank you very muach.

squeek502 (approved):

PR #565, opened 2021-09-16
fix #564, thread argument

### LINKED ISSUE #564: Crash when passing a lua object as thread argument
The main thread will crash if I pass a table, etc. as an argument to the thread, as shown below.

```
LuaJIT 2.1.0-beta3 -- Copyright (C) 2005-2021 Mike Pall. https://luajit.org/
JIT: ON SSE3 SSE4.1 BMI2 fold cse dce fwd dse narrow loop abc sink fuse
> uv=require'luv'
> t=uv.new_thread(function(tbl) print(tbl) end, {})
Error: thread arg not support type 'table' at 2> Error: thread arg not support type table at 1Uncaught Error: attempt to call a nil value
```

This problem can be fixed with the following patch.

```diff
diff --git a/src/thread.c b/src/thread.c
index b130593..50ed7e2 100644
--- a/src/thread.c
+++ b/src/thread.c
@@ -201,6 +201,7 @@ static int luv_thread_arg_push(lua_State* L, luv_thread_arg
_t* args, int flags)
     default:
       fprintf(stderr, "Error: thread arg not support type %s at %d",
         lua_typename(L, arg->type), i + 1);
+      lua_pushnil(L);
     }
     i++;
   };
```

However, I think it would be better to change the following part to `return lua_error(L)`. If we call `lua_error` with `luv_thread_arg_set`, will there be a memory leak or some other problem? I have not created a PR because I do not know which modification is preferable for us.
 
https://github.com/luvit/luv/blob/b2572ecf988185b9763965c590f6edbb2bc9b6c3/src/thread.c#L105-L110

zhaozg:


src/thread.c
```diff
@@ -103,8 +103,8 @@ static int luv_thread_arg_set(lua_State* L, luv_thread_arg_t* args, int idx, int
       }
       break;
     default:
-      fprintf(stderr, "Error: thread arg not support type '%s' at %d",
-        lua_typename(L, arg->type), i);
+      fprintf(stderr, "Error: thread arg #%d not support %s type\n",
```
bfredl, line 106: can we avoid `stderr` here as well? IIUC, `lua_error` would be ok here as there is a caller to return to, so it won't bring down the entire thread?

zhaozg: ok, I need more time to think and test, this api must support called safe in uv threadpools.

zhaozg:
More like https://github.com/luvit/luv/pull/566

PR #566, opened 2021-09-17
fix #564, invalid thread argument causes crash.

### LINKED ISSUE #564: Crash when passing a lua object as thread argument

erw7:
This is an attempt to fix #564 with a different approach than #565.

Change `luv_work_cb()` to run in protected mode and throw an error in `luv_thread_arg_set()` if an argument of an impermissible type is passed. It's a little more complicated, but I've integrated it into `neovim` and it seems to be working.

### DIFF
--- a/src/work.c
+++ b/src/work.c
@@ -226,6 +226,13 @@ static int luv_queue_work(lua_State* L) {
 
   ret = luv_thread_arg_set(L, &work->args, 2, top, LUVF_THREAD_SIDE_MAIN); //clear in sub threads,luv_work_cb
   if (ret < 0) {
+    //cache lua_State to reuse
+    lua_rawgeti(L, LUA_REGISTRYINDEX, ctx->pool_ref);
+    int i = lua_rawlen(L, -1);
+    *(lua_State**)lua_newuserdata(L, sizeof(lua_State*)) = work->args.L;
+    lua_rawseti(L, -2, i+1);
+    lua_pop(L, 1);
+
     luv_thread_arg_clear(L, &work->args, LUVF_THREAD_SIDE_MAIN);
     free(work);
     return luv_thread_arg_error(L);
@@ -234,6 +241,13 @@ static int luv_queue_work(lua_State* L) {
   work->work.data = work;
   ret = uv_queue_work(luv_loop(L), &work->work, luv_work_cb_wrapper, luv_after_work_cb);
   if (ret < 0) {
+    //cache lua_State to reuse
+    lua_rawgeti(L, LUA_REGISTRYINDEX, ctx->pool_ref);
+    int i = lua_rawlen(L, -1);
+    *(lua_State**)lua_newuserdata(L, sizeof(lua_State*)) = work->args.L;
+    lua_rawseti(L, -2, i+1);
+    lua_pop(L, 1);
+
     luv_thread_arg_clear(L, &work->args, LUVF_THREAD_SIDE_MAIN);
     free(work);
     return luv_error(L, ret);
--- a/src/work.c
+++ b/src/work.c
@@ -102,6 +102,9 @@ static int luv_work_cb(lua_State* L) {
       //clear in main threads, luv_after_work_cb
       i = luv_thread_arg_set(L, &work->rets, top + 1, lua_gettop(L),
           LUVF_THREAD_MODE_ASYNC|LUVF_THREAD_SIDE_CHILD);
+      if (i < 0) {
+        return luv_thread_arg_error(L);
+      }
       lua_pop(L, i);  // pop all returned value
       luv_thread_arg_clear(L, &work->rets, LUVF_THREAD_MODE_ASYNC|LUVF_THREAD_SIDE_CHILD);
     }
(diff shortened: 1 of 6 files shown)


zhaozg (approved):

zhaozg:
Would you like to add some test about this?

erw7:
> Would you like to add some test about this?

Yes, I would like to add a test. If I link to neovim, it looks like the problem has been solved. However, luajit still crash SIGABRT (core dumped). I'll give it some more consideration.

src/work.c
```diff
@@ -104,14 +107,30 @@ static void luv_work_cb(uv_work_t* req) {
...
+  luv_work_ctx_t* ctx = work->ctx;
+  lua_State *L = work->args.L;
+  luv_ctx_t* lctx = luv_context(L);
+
+  // If exit is called on a thread in the thread pool, abort is called in
+  // uv__threadpool_cleanup, so exit is not called in luv_cfpcall.
+  int i = lctx->thrd_cpcall(L, luv_work_cb, (void*)req, LUVF_CALLBACK_NOEXIT);
+  if (i != LUA_OK) {
+    luv_thread_arg_clear(L, &work->rets, LUVF_THREAD_MODE_ASYNC|LUVF_THREAD_SIDE_CHILD);
+    luv_thread_arg_clear(L, &work->args, LUVF_THREAD_SIDE_CHILD);
```
zhaozg, line 132: Add   ` free(work);` here should fix https://github.com/luvit/luv/pull/566/checks?check_run_id=3639647046#step:5:600

src/work.c
```diff
@@ -104,14 +107,30 @@ static void luv_work_cb(uv_work_t* req) {
...
+  luv_work_t* work =  (luv_work_t*)req->data;
+  luv_work_ctx_t* ctx = work->ctx;
+  lua_State *L = work->args.L;
+  luv_ctx_t* lctx = luv_context(L);
+
+  // If exit is called on a thread in the thread pool, abort is called in
+  // uv__threadpool_cleanup, so exit is not called in luv_cfpcall.
+  int i = lctx->thrd_cpcall(L, luv_work_cb, (void*)req, LUVF_CALLBACK_NOEXIT);
+  if (i != LUA_OK) {
+    luv_thread_arg_clear(L, &work->rets, LUVF_THREAD_MODE_ASYNC|LUVF_THREAD_SIDE_CHILD);
```
zhaozg, line 131: Lua vm must be push back to vm pools cache in main threads.
Should copy and refact code from 
https://github.com/luvit/luv/blob/b2572ecf988185b9763965c590f6edbb2bc9b6c3/src/work.c#L129-L138

Run here means not call luv_after_work_cb

zhaozg (changes requested):

src/work.c
```diff
@@ -104,14 +107,30 @@ static void luv_work_cb(uv_work_t* req) {
...
+static void luv_work_cb_wrapper(uv_work_t* req) {
+  luv_work_t* work =  (luv_work_t*)req->data;
+  luv_work_ctx_t* ctx = work->ctx;
+  lua_State *L = work->args.L;
+  luv_ctx_t* lctx = luv_context(L);
+
+  // If exit is called on a thread in the thread pool, abort is called in
+  // uv__threadpool_cleanup, so exit is not called in luv_cfpcall.
+  int i = lctx->thrd_cpcall(L, luv_work_cb, (void*)req, LUVF_CALLBACK_NOEXIT);
+  if (i != LUA_OK) {
```
zhaozg, line 130: Lua vm must be push back to vm pools cache in main threads to fix https://github.com/luvit/luv/pull/566/checks?check_run_id=3639646912#step:4:9914

Maybe we can copy and refact code from 
https://github.com/luvit/luv/blob/b2572ecf988185b9763965c590f6edbb2bc9b6c3/src/work.c#L129-L138

Run here means not call luv_after_work_cb

zhaozg (changes requested):

src/work.c
zhaozg: Or do a wrapper for luv_work_cb, insure luv_after_work_cb called everytime is a good idea.

src/work.c
zhaozg: Or do a wrapper for luv_work_cb, insure luv_after_work_cb called everytime is a good idea.

src/work.c
```diff
@@ -202,11 +224,17 @@ static int luv_queue_work(lua_State* L) {
     work->args.L = acquire_vm_cb();
   lua_pop(L, 1);
 
-  luv_thread_arg_set(L, &work->args, 2, top, LUVF_THREAD_SIDE_MAIN); //clear in sub threads,luv_work_cb
+  ret = luv_thread_arg_set(L, &work->args, 2, top, LUVF_THREAD_SIDE_MAIN); //clear in sub threads,luv_work_cb
+  if (ret < 0) {
+    luv_thread_arg_clear(L, &work->args, LUVF_THREAD_SIDE_MAIN);
```
zhaozg, line 229: https://github.com/luvit/luv/blob/b2572ecf988185b9763965c590f6edbb2bc9b6c3/src/work.c#L129-L134

zhaozg: This should help

erw7: @zhaozg  thanks.

zhaozg:
@erw7 Thank you very much.

PR #578, opened 2022-01-05
Update to Libuv 1.43.0

squeek502:
Only new function is uv_ip_name which is a generic function to convert a binary ip4/ip6 address to a string. We only used the protocol-specific versions of this function (uv_ip4_name/uv_ip6_name) internally, and it doesn't look like we have a use for the new function internally, so it's been left unbound and added to the list of unbound (not ruled out) functions in `bindcov.sh`.

From https://github.com/libuv/libuv/issues/3397:

> Notable changes:
> * `uv_ip_name()` has been added.
> * On Windows, `uv_fs_read()` and `uv_fs_write()` now return `UV_EBADF` instead of `UV_EPERM` when the file descriptor was opened with incorrect flags. This matches the behavior on other platforms.
> * Streams can be reset and attempted to read from after EOF is read.
> * The main thread now has a valid ID instead of `NULL`.
> * `uv_cpu_info()` now supports `aarch64`.
> * Several minor issues related to mingw32 have been fixed.

Full changelog: https://github.com/libuv/libuv/blob/v1.x/ChangeLog


zhaozg (approved):
So timely, you are number one.

PR #593, opened 2022-03-24
Add `luarocks test` support

### LINKED ISSUE #592: specify how to run tests in the rockspec
hi,

I am trying to bump the luv package in nixos https://github.com/NixOS/nixpkgs/issues/163134 to resume development on neovim. I found some issues in the nix packaging and I thought I would enable the luv testsuite. 
luarocks has had support for tests for a few years now, it would be nice to leverage it so that the user doesn't have to look at how to run tests but can just run `luarocks test`. It helps automating the maintainance of lua packages in nixos as we can default to `luarocks test` too. 

The doc is sparse on the wiki but here is a reference https://www.lua.org/wshop18/Muhammad.pdf (page 10) or even better the luarocks package itself.

squeek502:
I think this might drop support for Luarocks earlier than 3.0, but not totally sure. Closes #592

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,17 +84,13 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest]
-        luarocks_version: [2.4.4, 3.8.0]
+        luarocks_version: [3.8.0]
         lua_engine: [Lua, LuaJIT]
         include:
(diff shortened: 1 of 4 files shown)


zhaozg (approved):
Luarocks v2.4.4 is 4 years olds.

zhaozg:
Please feel free to merge.

PR #596, opened 2022-05-02
optimize luv_thread_dumped

zhaozg:


src/thread.c
```diff
@@ -219,34 +219,30 @@ static int luv_thread_arg_error(lua_State *L) {
 // luaL_buffinit might push stuff onto the stack (this is undocumented as of now),
 // so it must be called after lua_dump to ensure that the function to dump
 // is still on the top of the stack
```
squeek502, line 221: This comment (including the 'Copied from' line above here) should be removed.

src/thread.c
```diff
@@ -219,34 +219,30 @@ static int luv_thread_arg_error(lua_State *L) {
...
   if (lua_isstring(L, idx)) {
     lua_pushvalue(L, idx);
   } else {
-    int ret;
-    struct luv_thread_Writer state;
-    state.init = 0;
+    int ret, top;
+    luaL_Buffer B;
+
+    luaL_buffinit(L, &B);
```
squeek502, line 238:
```suggestion
    // In Lua >= 5.4.3, luaL_buffinit pushes a value onto the stack, so it needs to be called
    // here to ensure that the function is at the top of the stack during the lua_dump call
    luaL_buffinit(L, &B);
```

squeek502 (changes requested):

src/thread.c
zhaozg: That's clear

squeek502 (approved):

PR #604, opened 2022-06-01
fix #603

### LINKED ISSUE #603: Segmentation fault sometimes occurs on latest Neovim
I'm not quite sure under what conditions this happens, but it seems that sometimes `req->data` in the following code is NULL.
https://github.com/luvit/luv/blob/02d703b42be44d31483582a6538e885fa9b37134/src/fs.c#L38

As a result, Neovim falls down with the segmentation fault.
This occurs from commit 02d703b42be44d31483582a6538e885fa9b37134. Before that, it did not reproduce.

I would like to add a NULL check at the beginning of the following function, what do you think?

https://github.com/luvit/luv/blob/02d703b42be44d31483582a6538e885fa9b37134/src/lreq.c#L70
```c
  if (data == NULL) {
    return;
  }
```

ref: https://github.com/luvit/luv/issues/600

zhaozg:


squeek502:
As mentioned in #603, I think it'd be best to reproduce the bug before attempting a fix.

zhaozg:
> As mentioned in #603, I think it'd be best to reproduce the bug before attempting a fix.

Ok, It's your turn.

PR #613, opened 2022-09-06
add zig build system for luv

truemedian:
This is the first step in allowing simple cross compilation for luv and luvi via zig.

This code needs to be compared against the respective Makefiles and CMakeLists before this can be merged.

This is marked as a draft PR as some features (specifically tests) need to be able to pass successfully on all the supported platforms before this can be any semblance of a good way to build luv.

At the moment:
`tests/test-signal`'s SIGINT test needed to be modified to run `os.exit(0)`, this needs to be checked,
~~the built `luajit` executable is being flagged by UBSan~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 2.8.12)
 
 if(POLICY CMP0053)
   cmake_policy(SET CMP0053 NEW) # faster evaluation of variable references


squeek502:
> the built `luajit` executable is being flagged by UBSan

I assume this is being worked around via `disable_sanitize_c = true`? Or is it still triggering UBSan checks with that?

truemedian:
> I assume this is being worked around via disable_sanitize_c = true?

It *should* be, but yes, its still emitting `ub1` instructions and I didn't take the time to figure out why.

> Seems like it would be nice to try adding some more architectures to CI

I'll play around a bit with making github actions workflows for some of the major ones (linux64, macos64, win64/32) once I can get it all working correctly and passing tests.

truemedian:
I figured out why luajit was being ubsan'd (the library itself needed no-sanitize-c), however, something in libc's atexit is causing a segfault, and I can't figure out why that would be happening. It seems to be happening with both luajit and PUC, so its likely an issue with building libuv.

PR #618, opened 2022-09-26
Allow Coroutine Continuations

truemedian:
Allows passing a coroutine in place of a callback in all methods using uv_req_t. The coroutine will be yielded and then resumed using the synchronous argument list [not the (err, value) callbacks recieve]. I have not yet completely verified the behaviour when passing a coroutine that is not the running thread, however it should work.

Disabled by default, any blocking function call inside of the not-main coroutine will be a yield + resume to allow the libuv event loop to continue running. This behavior can be changed with the LUV_FORCE_COROUTINE_CONTINUATION flag to cmake.

Tap has been modified to *also* run tests inside of a coroutine to provide a testing suite for these changes.

I have not run into any issues so far, but the luv test suite definitely does not cover all bases.


CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ if(NOT ${CMAKE_VERSION} VERSION_LESS "3.5.0")
   set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 endif()
 
+option(LUV_FORCE_COROUTINE_CONTINUATION "Build as module" OFF)
```
squeek502, line 19: Description needs to be updated

src/lreq.c
```diff
@@ -16,10 +16,28 @@
  */
 #include "private.h"
 
-
 static int luv_check_continuation(lua_State* L, int index) {
-  if (lua_isnoneornil(L, index)) return LUA_NOREF;
-  luv_check_callable(L, index);
+#ifdef LUV_FORCE_COROUTINE_CONTINUATION
+// Uses the current thread unless it is the main thread
+if (lua_isnoneornil(L, index)) {
```
squeek502, line 22: Indentation seems off here

squeek502 (commented):
Just some minor nitpicks for now, haven't looked at the implementation too much yet.

squeek502:
Just to help me wrap my head around this, could you provide some simple example code showing a use-case this change allows (e.g. before, code with callbacks would be written like *this*, but now it can be written like *this*).

zhaozg:
Should add some tests to cover with coroutine or not.

truemedian:
> Just to help me wrap my head around this, could you provide some example code showing a use-case this change allows for (e.g. before, code with callbacks would be written like _this_, but now it could be written like _this_).

A quick and simple comparison would be using uv_fs_ (since its the largest consumer of uv_req_t).

completely async:
```lua
uv.fs_open(path, flags, mode, function(err, fd)
  assert(fd, err)
  -- do something with fd
end
```

currently completely blocking: 
```lua
local fd, errstr, errname = uv.fs_open(path, flags, mode)
assert(fd, errstr)
-- do something with fd
```
note: this would be a coroutine continuation if inside a non-main coroutine with `LUV_FORCE_COROUTINE_CONTINUATION` defined

coroutine continuation (yield + resume):
```lua
local fd, errstr, errname = uv.fs_open(path, flags, mode, coroutine.running())
assert(fd, errstr)
-- do something with fd
```

I attempted to make the resume of the coroutine match a blocking function call so that LUV_FORCE_COROUTINE_CONTINUATION can work seamlessly.

> Should add some tests to cover with coroutine or not.

I definitely agree that this needs a lot more tests. Thats what adding coroutine.wrap()'d tests to tap was intended to accomplish; but it doesn't test the new surfaces that I introduced.

Bilal2453:
Seems like the added test is failing on LuaJIT

truemedian:
I can't quite understand why exactly the appveyor build is failing. The failing tcp test passes on my machine on windows and there seems to be garbage at the bottom of the log?

truemedian:
Closed because it makes continuations much more complicated. Similar behavior can be reproduced via:

```lua
local thread = coroutine.running()
local req, err = uv.write(stream, function(...)
  assert(coroutine.resume(thread, ...)) -- resume waiting coroutine, beware this will eat your stack trace
end
if not req then
  return req, err -- write failed for some reason
end
return coroutine.yield() -- wait for write to finish

PR #620, opened 2022-10-13
docs: document uv.errno

Bilal2453:
documents changes introduced in https://github.com/luvit/luv/commit/a817184ef450bc31702db8e03cefc7ee67878624

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -133,6 +133,7 @@ success, or sometimes nothing at all. These cases are documented below.
 ### `uv.errno`
 
 A table value which exposes error constants as a map, where the key is the error name (without the `UV_` prefix) and its value is a negative number.
+See Libuv's [Error constants](https://docs.libuv.org/en/v1.x/errors.html#error-constants) page for further details.
 
 - `E2BIG`: argument list too long.
 - `EACCES`: permission denied.
--- a/docs.md
+++ b/docs.md
@@ -39,8 +39,8 @@ uv.run() -- an explicit run call is necessary outside of luvit
 The luv library contains a single Lua module referred to hereafter as `uv` for
 simplicity. This module consists mostly of functions with names corresponding to
 their original libuv versions. For example, the libuv function `uv_tcp_bind` has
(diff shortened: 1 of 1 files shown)


docs.md
```diff
@@ -130,6 +130,91 @@ When a function is called successfully, it will return either a value that is
...
+- `ETXTBSY`: text file is busy.
+- `EXDEV`: cross-device link not permitted.
+- `UNKNOWN`: unknown error.
+- `EOF`: end of file.
+- `ENXIO`: no such device or address.
+- `EMLINK`: too many links.
+- `ENOTTY`: inappropriate ioctl for device.
+- `EFTYPE`: inappropriate file type or format.
+- `EILSEQ`: illegal byte sequence.
+- `ESOCKTNOSUPPORT`: socket type not supported.
```
squeek502, line 216: Would it be better to link out to https://docs.libuv.org/en/v1.x/errors.html#error-constants instead of enumerating them here? It would ensure that we wouldn't have to keep this list in sync.

squeek502 (commented):
Thanks for doing this!

docs.md
SinisterRectus: Possibly, if it's autogenerated. It's nice to see the list here though too.

squeek502: The `errno` table is populated using libuv's [`UV_ERRNO_MAP` macro](https://docs.libuv.org/en/v1.x/errors.html#c.UV_ERRNO_MAP), so it'll always match the errors of version we build against.

Bilal2453: I was -and still- debating whether we should just link that page or not.  What I thought was "enumerating them here ain't gonna hurt", I guess we could do both, a link to the libuv docs page, while also listing them here (kinda nice to have em here in my opinion).  

Does mean we will need to pay *some* attention to those values being updated, but to be honest that won't be happening that often, if at all.

squeek502: Sounds good, I think adding the link would be good.

squeek502 (approved):

PR #631, opened 2023-01-27
build: fix cmake find module names

dundargoc:
CMake currently gives a warning since the file name (FindLibuv) isn't
the same as the argument given to FIND_PACKAGE_HANDLE_STANDARD_ARGS
(LIBUV). Changing the argument given to be the same as the file fixes
the warning.


SinisterRectus (approved):
Tests pass. Looks good?

PR #634, opened 2023-03-02
fix segfault in luv_check_handle/luv_check_stream

### LINKED ISSUE #633: segfault on luv_check_handle
I've been playing a bit with `luv.handle_get_type`/`luv.fileno` and I noticed that on Lua 5.1/LuaJIT it segfaults with
```lua
luv.handle_get_type(newproxy())
-- or
luv.fileno(newproxy())
```

I traced this back a bit to find it happens in `luv_check_handle`, more specifically on 
```c
if (!handle->data) { goto fail; }
```

I assume this only really happens when passing an empty userdata like with `newproxy`? Why exactly it happens or how it could be possibly fixed is beyond me though, I speculated that this is trying to index the userdata memory block out of bounds but not entirely sure if that is how userdata works.

Bilal2453:
Fixes #633.

Is the style I used here OK? It felt kind of different than the rest of the code, the assignments in the if statements more specifically..

### DIFF
--- a/src/handle.c
+++ b/src/handle.c
@@ -32,9 +32,9 @@ static uv_handle_t* luv_check_handle(lua_State* L, int index) {
   int isHandle;
   uv_handle_t* handle;
   void *udata;
-  if (!(udata = lua_touserdata(L, index))) { goto fail; }
-  if (!(handle = *(uv_handle_t**) udata)) { goto fail; }
-  if (!handle->data) { goto fail; }
+  // check if input is a userdata
+  udata = lua_touserdata(L, index);
+  if (!udata) goto fail;
   // "uv_handle" in the registry is a table structured like so:
   // {
   //   [<uv_aync metatable>] = true,
@@ -45,11 +45,17 @@ static uv_handle_t* luv_check_handle(lua_State* L, int index) {
   // we get its metatable and check that we get `true` back
   // when looking the metatable up in the "uv_handle" table.
   lua_getfield(L, LUA_REGISTRYINDEX, "uv_handle");
-  lua_getmetatable(L, index < 0 ? index - 1 : index);
+  isHandle = lua_getmetatable(L, index < 0 ? index - 1 : index);
+  if (!isHandle) goto fail;
   lua_rawget(L, -2);
   isHandle = lua_toboolean(L, -1);
   lua_pop(L, 2);
-  if (isHandle) { return handle; }
+  if (!isHandle) goto fail;
+  // cast the uesrdata to uv_handle_t
+  handle = *(uv_handle_t**)udata;
+  if (!handle || !handle->data) goto fail;
+  return handle;
+
   fail: luaL_argerror(L, index, "Expected uv_handle userdata");
   return NULL;
 }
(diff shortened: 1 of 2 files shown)


src/handle.c
```diff
@@ -45,11 +45,17 @@ static uv_handle_t* luv_check_handle(lua_State* L, int index) {
...
+  isHandle = lua_getmetatable(L, index < 0 ? index - 1 : index);
+  if (!isHandle) goto fail;
   lua_rawget(L, -2);
   isHandle = lua_toboolean(L, -1);
   lua_pop(L, 2);
-  if (isHandle) { return handle; }
+  if (!isHandle) goto fail;
+  // cast the uesrdata to uv_handle_t
+  handle = *(uv_handle_t**)udata;
+  if (!handle || !handle->data) goto fail;
```
squeek502, line 56: Nitpick: I believe this `!handle` check is redundant, as the `*(uv_handle_t**)udata` cast cannot make `handle` null and we already know `udata` is not null.

```suggestion
  if (!handle->data) goto fail;
```

src/handle.c
```diff
@@ -45,11 +45,17 @@ static uv_handle_t* luv_check_handle(lua_State* L, int index) {
...
   lua_getfield(L, LUA_REGISTRYINDEX, "uv_handle");
-  lua_getmetatable(L, index < 0 ? index - 1 : index);
+  isHandle = lua_getmetatable(L, index < 0 ? index - 1 : index);
+  if (!isHandle) goto fail;
   lua_rawget(L, -2);
   isHandle = lua_toboolean(L, -1);
   lua_pop(L, 2);
-  if (isHandle) { return handle; }
+  if (!isHandle) goto fail;
+  // cast the uesrdata to uv_handle_t
```
squeek502, line 54:
```suggestion
  // cast the userdata to uv_handle_t
```

squeek502 (commented):
Some minor nitpicks, but overall looks good.

src/stream.c
```diff
@@ -33,11 +33,17 @@ static uv_stream_t* luv_check_stream(lua_State* L, int index) {
...
+  isStream = lua_getmetatable(L, index < 0 ? index - 1 : index);
+  if (!isStream) goto fail;
   lua_rawget(L, -2);
   isStream = lua_toboolean(L, -1);
   lua_pop(L, 2);
-  if (isStream) { return handle; }
+  if (!isStream) goto fail;
+  // cast the userdata to uv_stream_t
+  handle = *(uv_stream_t**)udata;
+  if (!handle || !handle->data) goto fail;
```
squeek502, line 44:
```suggestion
  if (!handle->data) goto fail;
```

Bilal2453: I should've commited the changes myself instead of using that Github commit suggestion thing, my bad

squeek502 (approved):

squeek502:
Thanks!

Bilal2453:
No worries, thank you for helping me out with this, still learning my way.
(Also sorry for that half-failed squash attempt on my end lol)

PR #635, opened 2023-03-03
docs: new_work & new_thread string argument

Bilal2453:
`new_thread` already had this documented, although the argument type was missing `or string`, it also implied that only bytecode is accepted which seems to be inaccurate, as Lua code is accepted too  (internally it just `load`).
The same applies to `new_work`.


squeek502:
Thanks!

squeek502 (approved):

PR #641, opened 2023-04-01
add some code quality tools

zhaozg:


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,79 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
+    paths-ignore:
+      - '/deps/'
```
squeek502, line 21:
```suggestion
      - 'deps'
```

It seems like `/deps/` isn't working judging by the alerts here: https://github.com/luvit/luv/pull/641#issuecomment-1492893744

squeek502 (approved):
It seems like we might want to disable getting CodeQL alerts for stuff in `deps` at some point but that's something we can figure out later.

PR #645, opened 2023-04-18
[docs] Correct parameter name and type of fs_lstat

jgottzen:
I guess this was a copy/paste mistake from fstat.


jgottzen:
My pleasure!

Issue #599, opened 2022-05-12
Crash during calling os.exit()


kevinhwang91:
`lua -v` output: `Lua 5.4.4  Copyright (C) 1994-2022 Lua.org, PUC-Rio`.

```lua
local uv = require('luv')

local function setTimeout(callback, ms)
    local timer = uv.new_timer()
    timer:start(ms, 0, function()
        timer:stop()
        timer:close()
        callback()
    end)
    return timer
end

setTimeout(function()
    os.exit(5, true)
end, 1000)

uv.run()
```

`coredumpctl debug`:

```log
gdb bt
#0  luv_close_cb (handle=0x55c11a041190) at /home/kevin/WorkSpaces/lua/luv/src/handle.c:80
#1  0x00007f7bb506d809 in uv__finish_close (handle=0x55c11a041190) at /home/kevin/WorkSpaces/lua/luv/deps/libuv/src/unix/core.c:308
#2  uv__run_closing_handles (loop=0x55c11a0439c8) at /home/kevin/WorkSpaces/lua/luv/deps/libuv/src/unix/core.c:322
#3  uv_run (loop=loop@entry=0x55c11a0439c8, mode=mode@entry=UV_RUN_DEFAULT) at /home/kevin/WorkSpaces/lua/luv/deps/libuv/src/unix/core.c:401
#4  0x00007f7bb505cba2 in loop_gc (L=<optimized out>) at /home/kevin/WorkSpaces/lua/luv/src/luv.c:788
#5  0x000055c118105cf6 in precallC (f=<optimized out>, nresults=<optimized out>, func=<optimized out>, L=<optimized out>) at ldo.c:506
#6  luaD_precall (L=L@entry=0x55c11a0362a8, func=0x55c11a036a10, nresults=0) at ldo.c:572
#7  0x000055c1181060f7 in ccall (inc=65537, nResults=<optimized out>, func=<optimized out>, L=0x55c11a0362a8) at ldo.c:607
#8  luaD_callnoyield (L=0x55c11a0362a8, func=<optimized out>, nResults=<optimized out>) at ldo.c:627
#9  0x000055c118104f1a in luaD_rawrunprotected (L=L@entry=0x55c11a0362a8, f=f@entry=0x55c118107770 <dothecall>, ud=ud@entry=0x0) at ldo.c:144
[...]

Issue #437, opened 2019-11-03
It's possible for Luv to call into Lua with an unusable state (i.e. during lua_close)


squeek502:
See #414 and #436.

This is a more general problem than I thought. There are two things here:

1. It's possible for Lua to be called out to during lua_close due to how handles work in Libuv. One example is in #414 where if there is a shutdown req on a stream handle during gc, then that gets called during uv_close of that handle (which can be during loop_gc). This can be triggered in other ways as well (like a close callback being called during loop_gc)
2. We aren't handling handle closing properly/safely. In our handle_gc function, we free the handle memory if it's closing, but that *is not safe to do*, because in Libuv CLOSING and CLOSED are separate states (there can still be pending things on a closing handle like the flag UV_HANDLE_ENDGAME_QUEUED). It's only safe to free a handle once the callback from uv_close is called.

https://github.com/luvit/luv/blob/ab4a9ad98f4afc290fb5885e7b72d6127044fc18/src/handle.c#L116-L127

Running this code will cause Luv to free the handle during GC and then Libuv will try to access it (because of UV_HANDLE_ENDGAME_QUEUED), causing an assert/crash/undefined behaviour (tested on Windows, different platforms have different closing procedures, so not sure if it happens everywhere):

```lua
local uv = require('luv')

local test = uv.new_pipe(false)
uv.close(test)
-- during loop_gc, test is marked as closing so luv frees its uv_handle_t memory,
-- but Libuv was not actually done with the handle and tries to access that memory
-- while the loop is walking the handles and closing them so that the loop can be closed
```

Note that adding a `uv.run()` at the end of the script fixes this specific case.

---

I'm not quite sure how to fix either of these. I'll need to do more research and maybe check into how Node handles this sort of thing if possible (Node might be different though since Luv gives direct access to uv_run and I don't think Node does?).

squeek502:
Partially addressed by https://github.com/luvit/luv/pull/654. See https://github.com/luvit/luv/pull/654#pullrequestreview-1576122356 for what's left to do for this issue.

PR #654, opened 2023-05-26
chore: fix crash in `loop_gc`

zhaozg:
Fix https://github.com/luvit/luv/issues/599
Fix part of https://github.com/luvit/luv/issues/437

### DIFF
--- a/src/handle.c
+++ b/src/handle.c
@@ -88,13 +88,19 @@ static int luv_is_closing(lua_State* L) {
   return 1;
 }
 
+static void luv_handle_free(uv_handle_t* handle);
+
 static void luv_close_cb(uv_handle_t* handle) {
   lua_State* L;
   luv_handle_t* data = (luv_handle_t*)handle->data;
   if (!data) return;
   L = data->ctx->L;
-  luv_call_callback(L, data, LUV_CLOSED, 0);
-  luv_unref_handle(L, data);
+  if(data->ref > 0) {
+    luv_call_callback(L, data, LUV_CLOSED, 0);
+    luv_unref_handle(L, data);
+  } else {
+    luv_handle_free(handle);
+  }
 }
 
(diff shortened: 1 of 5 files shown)


zhaozg:
So hard, long time, but a few code change.

We use ref avoid gc call before close_cb, but `lua_close` will  prioritize `userdata` gc , cause `reference` failed.
So crash/ signal 11 occures.

Now use `luv_handle-t->ref` as guard, to do right memory free. Please review, check and verify.

squeek502 (approved):
Nice work on this, sorry for taking so long to review it.

Note that this only fixes part of #437, though. It's still possible to get callback functions called during lua_close/GC, which can still lead to problems. For example:

```lua
local uv = require('luv')

uv.fs_opendir('.', function()
  print("opendir cb called")
end)

-- note: calling uv.run() would fix this example
```

will result in (with a `printf`' added to `loop_gc`):

```
loop_gc
opendir cb called
```

During that callback, it could attempt to use something that has been garbage collected (and potentially freed) so there's a high chance of errors/crashes/UB/etc.

For example, this will segfault/UAF for me on Linux because it will free the `tty` and then try to read it in the opendir callback:

```lua
local uv = require('luv')

local tty = uv.new_tty(1, false)

uv.fs_opendir('.', function()
  print("opendir cb called")
  print(tty)
end)

-- note: calling uv.run() would fix this example
```

with some more debug prints added:

```
handle_gc: 0x4bae640
loop_gc
handle_free: 0x4bae640
opendir cb called
==12330== Invalid read of size 8
==12330==    at 0x4F994F0: luv_check_handle.constprop.0 (handle.c:56)
==12330==    at 0x4F9973C: luv_handle_tostring (handle.c:65)
```

But this is a separate problem to the one solved in this PR, so it doesn't need to block this being merged. I've edited the OP to make sure this doesn't close #437.

PR #655, opened 2023-05-30
Add access(2) constants to the constants table

rpigott:
fs_access is equivalent to access(2), but this function is useless without the corresponsing access mode constants F_OK, R_OK, W_OK, and X_OK. Let's add these constants to the table.

### DIFF
--- a/src/constants.c
+++ b/src/constants.c
@@ -73,6 +73,22 @@ static int luv_constants(lua_State* L) {
   lua_pushinteger(L, O_TRUNC);
   lua_setfield(L, -2, "O_TRUNC");
 #endif
+#ifdef R_OK
+  lua_pushinteger(L, R_OK);
+  lua_setfield(L, -2, "R_OK");
+#endif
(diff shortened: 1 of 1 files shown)


squeek502 (approved):
Seems fine to add these, but note that the mode can be specified as a string:

https://github.com/luvit/luv/blob/16bbffd24dbe046981994906f4f0a2a958f563dd/docs.md#L2844-L2846
which is usually how it's used, e.g. [here](https://github.com/luvit/luvit/blob/a8c708e064cc1d60ecca8d8ca00bc28fa2e49eeb/deps/require.lua#L295)

Implementation is here for reference:

https://github.com/luvit/luv/blob/16bbffd24dbe046981994906f4f0a2a958f563dd/src/fs.c#L179-L207

squeek502:
Thanks!

rpigott:
> mode can be specified as a string

I didn't know that! TYVM for the info.

PR #666, opened 2023-08-20
docs: make process_kill signum arg optional

Bilal2453:
In Libuv, [uv_process_kill](https://docs.libuv.org/en/v1.x/process.html#c.uv_process_kill) takes a process and a signum, where the signum is required. In Luv although, in luv_parse_signal, if the input is not an integer and not a string, the default value of SIGTERM is used.

Effectively making signum an optional argument with the default string value `sigterm`.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -1279,7 +1279,7 @@ When the child process exits, `on_exit` is called with an exit code and signal.
 
 **Parameters:**
 - `process`: `uv_process_t userdata`
-- `signum`: `integer` or `string`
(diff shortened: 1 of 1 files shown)


squeek502:
The relevant code:

https://github.com/luvit/luv/blob/9e3c2320149884e5d809e9391f8580bb5f67bcdf/src/process.c#L302-L310

squeek502 (approved):

Bilal2453:
Seems like I forgot `uv.kill` in this PR.

Issue #673, opened 2023-09-02
How to pass command line string to uv.spawn？


linrongbin16:
Hi, I read the docs and it looks like it's mandatory to pass list of args to uv.spawn.

But if I have some command line strings:

```bash
git log --pretty='%C(yellow)%h %C(cyan)%cd %C(green)%aN%C(auto)%d %Creset%s' --date=short

rg --no-heading --column -- 'fzf'
```

How to pass the command string directly to uv.spawn?

I am writing a Neovim plugin with luv, so I don't have a library manager like luarock. All I have is Neovim environment.

I feel like parsing command line string is quite difficult, I cannot just split the string by whitespace.

squeek502:
Going from a string with quotes, etc to a list of arguments is typically handled by the shell, and each shell can do things differently (how to escape things, what needs to be escaped, etc), so there's no one *correct* way to do it AFAIK.

I believe what you'd want for those two command would be:

```lua
{
  "git",
  "log",
  "--pretty=%C(yellow)%h %C(cyan)%cd %C(green)%aN%C(auto)%d %Creset%s",
  "--date=short",
}
```
and
```lua
{
  "rg",
  "--no-heading",
  "--column",
  "--",
  "fzf",
}
```

linrongbin16:
Hi, thanks to confirm.

I have one more question, how does the quotes and escaping characters handled by uv.spawn?

So in your sample code, there's no single quote `'` around the `%C(yellow)%h %C(cyan)%cd %C(green)%aN%C(auto)%d %Creset%s`?

So inside uv.spawn, it will automatically add quotes for me?

squeek502:
Quotes are only necessary when determining if something is a single argument or multiple arguments. Once the arguments have been split, then quotes are no longer necessary. This splitting (and therefore removal of quotes) is what the shell usually does for you.

For example, if you run something like:
```
myprogram "some argument"
```
then in `myprogram` you'd get `some argument` (without the quotes) as argv[1].

So, when calling `uv.spawn`, you just call it with the arguments already split, so there's no need for quotes.

There is a caveat for Windows, though, since [it does actually transform it back into a single string before giving it to `CreateProcess`](https://github.com/luvit/luv/blob/master/docs.md#uvspawnpath-options-on_exit). See the `options.verbatim` flag if you need to do your own quoting for Windows:

[...]

PR #674, opened 2023-09-02
docs: Correct uv.spawn `options.args` docs about the first argument

squeek502:
This was left over from the libuv docs, but the luv bindings handle this differently and `options.args` should *only* include the command arguments, not the command name itself.

See https://github.com/luvit/luv/issues/673#issuecomment-1703780666

Example from our tests:

https://github.com/luvit/luv/blob/dcd1a1cad5b05634a7691402d6ca2f214fb4ae76/tests/test-process.lua#L14-L31

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -1235,10 +1235,11 @@ end)
 
 The `options` table accepts the following fields:
 
-  - `options.args` - Command line arguments as a list of string. The first
-  string should be the path to the program. On Windows, this uses CreateProcess
-  which concatenates the arguments into a string. This can cause some strange
-  errors. (See `options.verbatim` below for Windows.)
+  - `options.args` - Command line arguments as a list of strings. The first
+  string should *not* be the path to the program, since that is already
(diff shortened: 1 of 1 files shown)


zhaozg (approved):

docs.md
```diff
@@ -1235,10 +1235,11 @@ end)
 
 The `options` table accepts the following fields:
 
-  - `options.args` - Command line arguments as a list of string. The first
-  string should be the path to the program. On Windows, this uses CreateProcess
-  which concatenates the arguments into a string. This can cause some strange
-  errors. (See `options.verbatim` below for Windows.)
+  - `options.args` - Command line arguments as a list of strings. The first
+  string should *not* be the path to the program, since that is already
+  provided via `path`. On Windows, this uses CreateProcess which concatenates
```
zhaozg, line 1240: provided via `path` by `luv`, this is slightly different with `libuv`.

squeek502: I think it's okay to leave this out because these are the `luv` docs, so it should be implied.

PR #682, opened 2023-12-14
docs: new_async callback is not optional

Bilal2453:
From what I can see, the callback argument is not optional and can't be nil.

See https://github.com/luvit/luv/blob/master/src/async.c#L38.

I can't be 100% sure that the docs are wrong here and not the code itself for not allowing optional callbacks, but I can't see why would anyone need to create an async handle with no callback as that is counterintuitive.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -922,14 +922,14 @@ end)
 async:send()
 ```
 
-### `uv.new_async([callback])`
+### `uv.new_async(callback)`
 
 **Parameters:**
-- `callback`: `callable` or `nil`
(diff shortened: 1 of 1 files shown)


squeek502 (approved):
Thanks! Agreed about the `nil` callback.

Bilal2453:
oh sorry I forgot to mention, might need to take a look at the Windows version of those.  I don't think libuv will be returning an error in the Windows version but not the unix one, but I honestly didn't check the Windows ones.

PR #686, opened 2024-01-13
fix: Use os_uname() to check for Linux

cryptomilk:
In build environments like Fedora's mockbuild, there is no OSTYPE environment variable set. Use a more secure method to detect Linux. Details:

man os-release
https://www.man7.org/linux/man-pages/man5/os-release.5.html


zhaozg:
I think should use https://github.com/luvit/luv/blob/master/docs.md#uvos_uname

cryptomilk:
I've changed it.

cryptomilk:
The minimum supported libuv doesn't provide `os_uname()`. What now, fall back to `/etc/os-release`?

Bilal2453:
I don't think simply checking if `/etc/os-release` exists is as reliable either!  As far as I know, it is a systemd thing that may not exists on non-systemd distribution.  The most reliable way I know of for detecting if the OS is Linux based (works on Android environments as well) is by checking `uname --kernel-name` (`uname -s`), this will require an `io.popen` most likely, roughly something like this:
```lua
local popen_handle = io.popen('uname -s')
local uname_os = assert(popen_handle:read('*a'))
popen_handle:close()
isLinux = uname_os:lower() == 'linux'
```

I would also like to point out that in the current code it seems to me like `isLinux` is set to `nil` instead of `false` if the OS is not Windows but also isn't Linux (for example, when the OS is openbsd and `jit` isn't available).

cryptomilk:
I've updated it to use `uname -s`.

`/etc/os-release` is a freedesktop standard, it also has been added to Python 3.10, see https://docs.python.org/3/library/platform.html#platform.freedesktop_os_release

lib/utils.lua
```diff
@@ -19,10 +19,15 @@ if _G.jit and _G.jit.os then
 else
   -- Normal lua will only have \ for path separator on windows.
   utils.isWindows = package.config:find("\\") and true or false
+  utils.isLinux = false
+
   if not utils.isWindows then
-    local _os = os.getenv('RUNNER_OS') or os.getenv('OSTYPE')
-    if (_os and _os:lower():match('linux')) then
-      utils.isLinux = true
+    -- TODO Use uv.os_uname() when the minimum required libuv provides it
```
squeek502, line 25: There is `utils.uvVersionGEQ` defined in this file. Looks like `os_uname` was added in libuv 1.25.0, so the condition would be `utils.uvVersionGEQ("1.25.0")`

cryptomilk:
Looks like `popen()` is not alway available?

squeek502:
Looks like `popen` is dependent on [`LUA_USE_POSIX` being defined](https://github.com/lua/lua/blob/7923dbbf72da303ca1cca17efd24725668992f15/liolib.c#L57-L82), which we currently don't do. Adding

```cmake
IF(NOT WIN32)
  ADD_DEFINITIONS(-DLUA_USE_POSIX)
ENDIF(NOT WIN32)
```

to [`lua.cmake`](https://github.com/luvit/luv/blob/master/deps/lua.cmake) is probably a good idea in general and should fix the error.

Bilal2453:
@squeek502 I think the CI tests need to be manually re-triggered?

squeek502 (approved):
Thanks, this is a nice improvement to make the tests less dependent on the CI environment!

cryptomilk:
I've tested it with fedora mockbuild and the patch works fine.

PR #689, opened 2024-01-17
docs: typo in a table field name

Bilal2453:
the field name is `no_truncate` not `no_trunate`, it is correctly defined in the code.

I think the `true|false` format is also inconsistent, it will make things harder when generating the docs, should simply be something like `{ no_truncate: boolean }`, or the more consistent
```md
`flags`:
   `no_truncate`: `boolean`
```
update:
There is also a new `**Note**` format that was introduced with https://github.com/luvit/luv/commit/8e2f3066f9e999affa668f1dabe633e418b0cefd.
But maybe I will tackle those later when I progress with https://github.com/Bilal2453/luv-docgen.


squeek502 (approved):
Thanks!

PR #693, opened 2024-02-21
Annotate .gitmodules with branch and tag

creationix:
I'm working on an experiment to make submodules easier to work with.  Part of that is annotating .gitmodules with information to make it so tools can automatically update dependencies.

The `branch` field is actually supported already by [upstream git itself](https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt-set-branch-b--branchltbranchgt--ltpathgt).

The `tag` field is my invention and my tool may be using it to do semver updates based on the tag name and available upstream tags.

I'm not convinced the tag needs to be documented here since the commit hash is already in git and it possible to map that back to tags.  I've removed `tag` from the PR for now till I'm convinced it needs to be in the file.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,12 +1,18 @@
 [submodule "libuv"]
 	path = deps/libuv
 	url = https://github.com/libuv/libuv.git
+	branch = v1.x
+	tag = v1.48.0
 [submodule "luajit"]
 	path = deps/luajit
 	url = https://github.com/LuaJIT/LuaJIT.git
+	branch = v2.1
 [submodule "lua"]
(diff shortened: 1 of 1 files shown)


SinisterRectus:
Hi Tim! My 👍 just meant I approve of anything that makes the build process easier, but I don't know enough to properly sign off on this.

creationix:
No worries, thanks for the 👍🏽

squeek502 (approved):

Bilal2453:
This sounds interesting.  Depending on how it goes we might eventually make use of this in Lit as well!

PR #696, opened 2024-03-01
Fix garbage collection of scandir reqs

squeek502:
This is one possible approach to solving this. I'd like to test it a bit more and reproduce the crash in https://github.com/luvit/luv/pull/694#issuecomment-1973031055 to make sure this fixes it before merging.

---

Return a userdata wrapper around the uv_req_t to allow it to be garbage collected before the program exits. Previously, the returned userdata held a reference to itself in the Lua registry, meaning it would never be able to be garbage collected until the process ended.

This reverts commit 0e4a89515f385942b8f13322741cdbecad604d18, which attempted a workaround for the same underlying problem, but introduced a use-after-free.


zhaozg (approved):
well done

zhaozg:
reproduce before cmp with  https://github.com/FelipeLema/cmp-async-path.

image

squeek502:
Thanks! Minimal reproduction:

```lua
local uv = require('luv')

local entries = assert(uv.fs_scandir('.'))

local work
work = assert(uv.new_work(function(_entries)
  local _uv = require('luv')
  while true do
    if not _uv.fs_scandir_next(_entries) then break end
  end
end, function() end))

work:queue(entries)

uv.run()
```

PR #703, opened 2024-06-03
docs: Clarify types for mode arg in some filesystem functions

adigitoleo:
Some libuv filesystem functions expect the numeric `mode` argument to be an octal representation of the equivalent `chmod(1)` mode. This is potentially surprising behaviour to anyone who hasn't used either `libuv` or `glibc` from C. The documentation for e.g. `fs_mkdir` in libuv defers to `mkdir(2)` which defers to `mode_t(3)` which only says "It is an integer type". Furthermore, `fs_access` uses a different `mode` type, validated using
[luv_check_amode](https://github.com/luvit/luv/blob/7233e6dea92498a244feb51b790c1ba51e8abbff/src/fs.c#L184), which can accept strings ("symbolic mode" in `chmod(1)`). I chose not to mention that this one still technically accepts an integer because I have no idea about the restrictions of said integer.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -2666,7 +2666,7 @@ If `offset` is nil or omitted, it will default to `-1`, which indicates 'use and
 
 **Parameters:**
 - `path`: `string`
-- `mode`: `integer` (octal representation of `chmod(1)` mode)
+- `mode`: `integer` (octal `chmod(1)` mode, e.g. `tonumber('755', 8)`)
 - `callback`: `callable` (async version) or `nil` (sync version)
   - `err`: `nil` or `string`
   - `success`: `boolean` or `nil`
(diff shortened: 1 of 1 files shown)


docs.md
```diff
@@ -2591,7 +2591,7 @@ Equivalent to `close(2)`.
 **Parameters:**
 - `path`: `string`
 - `flags`: `string` or `integer`
-- `mode`: `integer`
+- `mode`: `integer` (octal representation of `chmod(1)` mode)
```
squeek502, line 2594: How about including an example like this:

```suggestion
- `mode`: `integer` (octal `chmod(1)` mode, e.g. `tonumber('644', 8)`)
```

(this is how the mode parameter is typically constructed in [the tests](https://github.com/luvit/luv/blob/7233e6dea92498a244feb51b790c1ba51e8abbff/tests/test-fs.lua#L4))

adigitoleo: If you want I can add it in the other places as well? Let me know.

squeek502: Yes, please do.

adigitoleo:
Added example usages as per review comment. `755` for mkdir instead of `644` (that is the usual directory permission code on Linux).

PR #708, opened 2024-07-06
Support passing tables and closures to new threads

cshuaimin:
Hi luvit team, first thank you for this great project! Luv makes it possible for calling libuv functions in Neovim!

In this PR I changed `thread_arg_*` functions to support `table` and `function` types to `new_thread()`, `queue_work()` and `async_send()` functions.

Althrough in some feature requesting issues you said that not supporting tables is intentional (I can't find the exact reply anymore), I believe it's a very important feature to have. It will unlock possibilities like multi-threaded Neovim plugins, off-loading some CPU intensive work to a thread pool. I sincerely invite you to re-think this feature.

I mainly translated Neovim's [deepcopy()](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/shared.lua#L13-L43) function to C. It uses a cache table to properly support reused table fields.

The function type supports upvalues. Supporting functions is needed because it allows Lua OOP values to be passed to new threads.

Testing: I added test and passed valgrind, and I've tested this branch in Neovim. However not tested in Lua versions other than LuaJIT.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -73,7 +73,7 @@ used here to facilitate documenting consistent behavior:
   metamethod
 - `buffer`: a `string` or a sequential `table` of `string`s
 - `threadargs`: variable arguments (`...`) of type `nil`, `boolean`, `number`,
-  `string`, or `userdata`, numbers of argument limited to 9.
+  `string`, `userdata`, `function`, or `table`, numbers of argument limited to 9.
 
 ## Contents
 
--- a/src/lthreadpool.h
+++ b/src/lthreadpool.h
@@ -21,8 +21,10 @@
 
 #define LUV_THREAD_MAXNUM_ARG 9
 
+typedef struct luv_table luv_table_t;
+
 typedef struct {
-  // support basic lua type LUA_TNIL, LUA_TBOOLEAN, LUA_TNUMBER, LUA_TSTRING
+  // support lua types of LUA_TNIL, LUA_TBOOLEAN, LUA_TNUMBER, LUA_TSTRING, LUA_TTABLE
   // and support uv_handle_t userdata
   int type;
   union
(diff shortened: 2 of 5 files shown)


cshuaimin:
I totally understand your concerns. I'm not very sure whether this is the best solution too. However in order to use the full power of CPU cores, in my opinion this complexity is unavoidable.

I have a nvim plugin that searches a file with tree-sitter. It can only search one file. To support searching multiple files under a directory, the plugin needs to parse hundreds of files with tree-sitter, which may block nvim UI for several minutes. If I can pass tables and functions (actually the [vim.treesitter.LanguageTree](https://github.com/neovim/neovim/blob/6a886a2511bbfd24a4d6ecc3f3a75f08a6df9de9/runtime/lua/vim/treesitter/languagetree.lua#L91) object) to Luv threads, then I can move parsing off main thread. Lua coroutine does not help here, because it's pure CPU work.

Aside from my own plugin, nvim's tree-sitter highlighting also runs on main thread. Opening or eding a large file with lots of tree-sitter injections war quite slow. Popular nvim distros disable tree-sitter highlights in large files. Nvim devs has already made this faster. With this PR it may be possible to make nvim UI even more responsive.

cshuaimin:
[Here](https://github.com/cshuaimin/ssr.nvim/blob/dev/lua/ssr/search.lua#L207-L255) is the relevant code in my plugin if you are curious. It's still experimental code.

truemedian:
> in my opinion this complexity is unavoidable.

This complexity is certainly avoidable in Luv. The solution you've implemented here is nothing more than a serialization and de-serialization of both tables and functions types.

Nothing is stopping you from doing this serialization *outside* of luv, potentially even serializing the respective table to a string entirely within Lua.

In my opinion this change is unnecessary and adds even more confusion around how luv threads work, because tables and functions are passed through the thread boundary by copy instead of by reference, which is the opposite of how the types behave normally.

From a user-facing perspective, this invisible copy of the types introduces the wrong behavior (eg. changing a table inside the thread does not change the table outside the thread; functions with upvalues behave strangely, because all upvalues are copies not references to their original values).

The most important problem I'd like to bring up is that this introduces incorrect and invalid behavior when serializing *any* function associated with a C library; most libraries are *not* thread-safe, and without a significant amount of added effort userdata will not get the correct metatable because the library was never required in the new Lua state and the metatables will not be present in the registry.

Bilal2453:
I am not in favor of adding this into Luv, I also believe it is an avoidable complexity that is better off being handled in Lua.  But also admittedly that won't be as straight forward, and as mentioned handling userdata on the Lua side is going to be tricky to say the least.

But also threading in Luv (and Lua generally) is really lacking, and just barely works, personally I think due to the design of Lua it is really unfavorable to do any extensive threading besides crushing some numbers/strings, so any work on that side is appreciated. 

How about adding support for this but outside of `new_thread(), queue_work(), async_send()`. For example adding new `new_thread_copy`, `queue_work_copy`, `async_send_copy`.  This will hopefully make it very clear what it is doing with the data to not confuse the user, and at the same time it isolates the unwanted complexity into its own experimental space. Will also ensure that we don't accidentally break already working code.

PR #723, opened 2024-10-19
Avoid SEGFAULTs in uv.walk when processing invalid handles

### LINKED ISSUE #721: Sharing the event loop leads to issues with processing foreign handles/cleanup
Following-up with https://github.com/luvit/luv/issues/718#issuecomment-2407216958):

The SEGFAULTs mentioned here are related to a specific use case, which seemingly violates a core assumption in the luv code. 

Handles ( `uv_handle_t`) managed by luv are allocated in `luv_setup_handle`, where they store the `luv_handle_t` as userdata:

https://github.com/luvit/luv/blob/5eb6543d9df0e9965b1f559c528473dcb4ce5003/src/lhandle.c#L31

https://github.com/luvit/luv/blob/5eb6543d9df0e9965b1f559c528473dcb4ce5003/src/lhandle.c#L50-L55

This works great when all handles are created in Lua and luv is the only user of the event loop. But when an application that embeds luv wants to integrate other libraries (that also support libuv) and then shares the `uv_loop_t` with luv, things can get a little dicey:

* luv code was clearly written under the assumption that all handles have a  `luv_handle_t` referenced by `handle->data`
* This assumptions doesn't hold when the handle was created externally (`data`  may be `NULL` or completely unrelated)
* During `uv.walk(),` `luv_walk_cb` reads the `data` member to access the Lua registry based on the stored `data->ref` 
* On shutdown, `loop_gc` runs the loop until all handles are closed, again assuming they were created by luv (and need closing)
* There are probably other scenarios where requests or handles might be created elsewhere and put on the same loop

The application might want to use libuv APIs directly (from C and/or the FFI), so a high degree of interoperability would be valuable. Beause `luv_loop()` is a public API I assumed that making use of the existing event loop was fine, even in the same Lua state. Some popular libraries that support sharing event loops would be [libwebsockets](https://github.com/warmcat/libwebsockets) and [uSockets](https://github.com/uNetworking/uSockets) (used in [uWebSockets](https://github.com/uNetworking/uWebSockets)). The latter is what I use.

I encountered the problem because the library [creates several handles](https://github.com/uNetworking/uSockets/blob/89db1704ddb1ece2aa447f550c863ba5524e0e4d/src/eventing/libuv.c#L134-L161) that trigger the assertion/crash when processed by luv:

https://github.com/luvit/luv/blob/5eb6543d9df0e9965b1f559c528473dcb4ce5003/src/loop.c#L92-L98

Here the foreign handles were being processed as if their `data`  pointed to a `luv_handle_t` reference, which it did not. I've since experimented with workarounds that allow skipping the irrelevant handles, but the lifetime/GC logic still choked on them. 

I'm not sure if this is a supported use case or if you'd rather see embedders adopt a different approach. Potential solutions:

* The application could create and manage multiple event loops, at the cost of increased complexity/unexpected behavior
* The application might create multiple Lua states and retrieve a separate `luv_loop` for each, again increasing complexity
* luv could tag the handles that it manages and then skip those created externally, at the cost of a larger footprint for all users
* luv could provide hooks that optionally allow the application to handle problematic aspects itself, increasing complexity
* luv could attempt to use the Lua registry instead of tagging handles, but this didn't seem to work reliably in my testing

There could be better ways of handling this. Or maybe the library already supports sharing the loop, but I didn't see how. Any ideas?

rdw-software:
Much ado for nothing, but this does resolve #721. Let me know if you prefer a different solution!

### DIFF
--- a/src/loop.c
+++ b/src/loop.c
@@ -95,7 +95,7 @@ static void luv_walk_cb(uv_handle_t* handle, void* arg) {
 
   // Sanity check
   // Most invalid values are large and refs are small, 0x1000000 is arbitrary.
-  assert(data && data->ref < 0x1000000);
+  if(!data || data->ref >= 0x1000000) return;
 
   lua_pushvalue(L, 1);           // Copy the function
   luv_find_handle(L, data);      // Get the userdata


src/loop.c
```diff
@@ -95,7 +95,7 @@ static void luv_walk_cb(uv_handle_t* handle, void* arg) {
 
   // Sanity check
   // Most invalid values are large and refs are small, 0x1000000 is arbitrary.
-  assert(data && data->ref < 0x1000000);
+  if(!data || data->ref >= 0x1000000) return;
```
zhaozg, line 98: The intent here is to ensure that the handles are created by Luv. Since the support for external (non-Luv-created) objects was not comprehensively considered in the original design of Luv, and there is no well-established mechanism to distinguish between external and internal objects, there are likely still issues present here.

Bilal2453: > there are likely still issues present here.

Should the solution instead tag all luv created handles, and check for the existence of that tag?  Simply making this a return seems fair to me

rdw-software: Yeah, I understand. What's why I proposed tagging external handles with magic values to make luv exit early here. That'd be a NO-OP for luv tests and you can leave the assertion unchanged. If there are other issues  only relevant to this use case, I'll probably find them sooner or later (with automated testing). Or indeed, could tag ALL luv handles.

Bilal2453: I believe the intent here should be simply to not crash unexpectedly for the end-user over something they might not have control for in production, rather than guaranteeing they are getting luv handles, which I would argue is out of scope for this PR and can be done later on, perhaps by as suggested tagging all luv handles.

zhaozg: It's hard to say; I don't have a safe and efficient solution at the moment. 

some options 

1.  tag, that with high performance, but not enough safe.
2. cache all data internal object pointer in a registry table, check it internal or external object,  that safe enough and and good performance.
3. other solution...

> > there are likely still issues present here.
> 
> Should the solution instead tag all luv created handles, and check for the existence of that tag? Simply making this a return seems fair to me

zhaozg:
Could you help me to test https://github.com/luvit/luv/pull/733, It's use solution.

> cache all data internal object pointer in a registry table, check it internal or external object, that safe enough and and good performance.

rdw-software:
> Could you help me to test #733, It's use solution.
> 
> > cache all data internal object pointer in a registry table, check it internal or external object, that safe enough and and good performance.

Your solution seems to work fine. My test cases no longer crash when applying the patch. Thanks!

rdw-software:
Obsolete since a reference to all luv-managed handles is now stored in the Lua registry.

Issue #505, opened 2020-07-07
memory leak related to `uv.send_async`


FelipeLema:
I made an ad-hoc by-line process output reader and it seems that I found a memory leak.

I wrapped into a min working example below. This is supposed to `uv.async_send` the _lines_ of the output of a certain process. I setup a process that prints data endlessly. The output is stored in `current_chunk`, but it's only temporary: each time data arrives, I look at the chunk and extract the lines, expect the last one (in case the last line did not arrive completely).

If you keep the `uv.send_async` line below, memory consumption of the `lua` process will grow endlessly. If you remove it, the memory consumption remains still. In both cases, each line is discarded.

```lua
local uv = require("luv") -- "luv" when stand-alone, "uv" in luvi apps-- WIP luv (related to luvit)

local stdout = uv.new_pipe()
local handle, pid =
   uv.spawn(
      "bash",
      {stdio = {nil, stdout, nil},
       args = {"-c",
               "while true; do ps aux; done"}
      },
      function(code,signal)
         -- ignore
      end)

local current_chunk  = "" -- accumulated data until a \n
local no_more_output = false
-- process each line
cb = uv.new_async(function(line)
      -- pass
      if no_more_output then
         cb:close()
      end
end)

uv.read_start(stdout, function(err, data)
                 assert(not err, err)
                 if data then
                    -- append received data into current_chunk
                    current_chunk = string.format("%s%s", current_chunk, data)
                 else
                    no_more_output= true
                 end

                 -- split chunk into lines to send to `callback`
                 lines={}
                 for line in string.gmatch(current_chunk, "[^\n]+") do
                       table.insert(lines, line)
                    end
                    if not no_more_output then
                       -- pop last chunk back into current chunk for next iteration
                       last_item_position = #lines
                       current_chunk = table.remove(lines, last_item_position)
                    end 

                    -- send the lines to the callback
                    for _,l in pairs(lines) do
                       -- this line below creates a memory leak
                       uv.async_send(cb, l)
                    end

                    -- close uv stuff if everything's done
                    if no_more_output then
                       uv.async_send(cb, nil)
                    end

end)
uv.run()
```

Happens with these lua versions:
```
% lua -v
Lua 5.3.3  Copyright (C) 1994-2016 Lua.org, PUC-Rio
% luajit -v
LuaJIT 2.1.0-beta3 -- Copyright (C) 2005-2017 Mike Pall. http://luajit.org/
```

squeek502:
[...]

PR #725, opened 2024-10-25
Protect async send from misuse

javalikescript:
This is a proposal to mitigate the async send issues, see #505
⚠️This is a breaking change as some misuse could work but with unexpected behavior.

Using arguments when calling async send is helpful but does not match the libuv implementation that will coalesce calls.
Properly supporting multiple calls is a lot of work and may not be interesting for luv.

I propose to restrict the usage by allowing arguments only when async send is used just once.
It is still possible to call multiple times async send without arguments which is the libuv API.
It seems to me to be a simple but necessary fix.

Another option would be to use a lock to be able to change the arguments and free the not yet received ones, but there will be no guaranty on which arguments will be effectively received. This option is only interesting to avoid breaking existing misuse.


javalikescript:
Closing in favor of #729

PR #727, opened 2024-11-02
Support Lua integer in thread and async arguments

javalikescript:
When passing integers, available since Lua 5.3, as thread or async arguments, the value pushed in the thread or async function is a float.

This change adds the integer flavor to the existing Lua number argument.

### DIFF
--- a/tests/test-thread.lua
+++ b/tests/test-thread.lua
@@ -62,14 +62,16 @@ return require('lib/tap')(function (test)
     local delay = 100
     uv.update_time()
     local before = uv.now()
-    local args = {delay, 'string', nil, false, 5, "helloworld"}
+    local args = {delay, 'string', nil, false, 5, 3.14, "helloworld"}
     local unpack = unpack or table.unpack
-    uv.new_thread({stack_size=0}, function(delay,s,null,bool,five,hw)
+    uv.new_thread({stack_size=0}, function(delay,s,null,bool,five,pi,hw)
       assert(type(delay) == "number")
       assert(type(s) == "string")
       assert(null == nil)
       assert(bool == false)
       assert(five == 5)
+      assert(math.type(five) == 'integer')
+      assert(pi == 3.14)
       assert(hw == 'helloworld')
       require('luv').sleep(delay)
     end, unpack(args)):join()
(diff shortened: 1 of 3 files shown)


tests/test-thread.lua
```diff
@@ -62,14 +62,16 @@ return require('lib/tap')(function (test)
...
+    local args = {delay, 'string', nil, false, 5, 3.14, "helloworld"}
     local unpack = unpack or table.unpack
-    uv.new_thread({stack_size=0}, function(delay,s,null,bool,five,hw)
+    uv.new_thread({stack_size=0}, function(delay,s,null,bool,five,pi,hw)
       assert(type(delay) == "number")
       assert(type(s) == "string")
       assert(null == nil)
       assert(bool == false)
       assert(five == 5)
+      assert(math.type(five) == 'integer')
```
squeek502, line 73:
```suggestion
      if math.type then
        assert(math.type(five) == 'integer')
      end
```

javalikescript: Fixed in commit 0d141bff9dcb3f05c2e19b40af48cd9f4448e8db

zhaozg (approved):
LGTM. The CI [errors](https://github.com/luvit/luv/actions/runs/11649845591/job/32437850489?pr=727) are irrelevant to this PR.

squeek502:
The CI failure is due to this assertion failure:

```
lua: /home/runner/work/luv/luv/src/loop.c:98: luv_walk_cb: Assertion `data && data->ref < 0x1000000' failed.
```

This is the same assertion discussed in https://github.com/luvit/luv/issues/721. Any idea why it would be getting triggered here?

Bilal2453:
Sounds like a good time to also look into https://github.com/luvit/luv/pull/723.

zhaozg:
Rerun fail ci, pass https://github.com/luvit/luv/actions/runs/11649845591/job/32524466086?pr=727.

squeek502:
That assertion ever failing in our own tests seems like a bad sign (and the changes in https://github.com/luvit/luv/pull/723 would potentially make this a silent failure).

I think it's unrelated to the changes in this PR, though.

PR #728, opened 2024-11-03
Use a special metatable for thread userdata arguments

javalikescript:
This is a proposal for handling thread and async userdata arguments.

The current implementation copies the userdata block of raw memory, sets the same metatable then removes it after the callback.
This implementation requires the userdata usage to be thread-safe and not used after the callback.

As far as I see the main use case for passing userdata consists in passing async to the thread callback.
It is not possible to check if a userdata is thread-safe.
The userdata cannot be used after the callback which is fine for thread but not for async.

This change consists in letting the caller opt-in to pass the metatable by providing a dedicated metatable suffixed `_ref`.
Such metatable is provided for `uv_async` with no close nor gc.
There is still a major issue as the async could be garbage collected on the main side.

Another option would be to forbid all userdata arguments except async or not setting the metatable at all.

### DIFF
--- a/src/async.c
+++ b/src/async.c
@@ -17,7 +17,12 @@
 #include "private.h"
 
 static uv_async_t* luv_check_async(lua_State* L, int index) {
-  uv_async_t* handle = (uv_async_t*)luv_checkudata(L, index, "uv_async");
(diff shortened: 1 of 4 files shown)


javalikescript:
Closed in favor of #729

PR #729, opened 2024-11-04
Fix async

### LINKED ISSUE #505: memory leak related to `uv.send_async`

javalikescript:
Adds a reference count to async enabling to pass it properly as thread/async argument.
Protects async arg using a mutex and defer clear to GC.
Copy async arguments rather than using Lua ref.

Fix #505
Fix passing an async as async send argument then use it outside the callback function
Fix suspect pop, [here](https://github.com/luvit/luv/blob/7233e6dea92498a244feb51b790c1ba51e8abbff/src/thread.c#L157)
Fix thread [join](https://github.com/luvit/luv/blob/7233e6dea92498a244feb51b790c1ba51e8abbff/src/thread.c#L488) after [exited](https://github.com/luvit/luv/blob/7233e6dea92498a244feb51b790c1ba51e8abbff/src/thread.c#L301)

This PR attempts to keep the async/thread behavior, except for setting the userdata metadata in the async callback and passing empty userdata as light userdata.
It may not be the best solution as it results in rather conplex and messy code.


javalikescript:
Using a reference count is not enough, the async handle from another thread should have a specific behavior such as dedicated GC and close methods.
I put this fix on hold looking for a simpler solution.

PR #733, opened 2024-11-09
try safe handle external handle

zhaozg:


zhaozg:
A random ci fail: https://github.com/zhaozg/luv/actions/runs/11754015398/job/32747531427

rdw-software (commented):
Didn't measure performance as it's unlikely to matter, but the loop can be shared without crashing when using the patch :)

zhaozg:
> Didn't measure performance as it's unlikely to matter, but the loop can be shared without crashing when using the patch :)

Thanks for your check, I'll update when I have free time, in this week.

clason:
Sorry for necro-posting, but can you comment on why these tests are built unconditionally? They break Neovim's build on Ubuntu (clang) and Windows: https://github.com/neovim/neovim/actions/runs/12924924457/job/36044861041?pr=32174

squeek502:
> but can you comment on why these tests are built unconditionally?

Just an oversight, sorry. Will be fixed by https://github.com/luvit/luv/pull/752

Issue #503, opened 2020-06-30
Incorrect references/usage of `L` causing invalid memory access


mwild1:
In a number of places luv stores a long-lived reference to `L` such as [src/luv.c:712](https://github.com/luvit/luv/blob/master/src/luv.c#L712).

This is not safe, however - `L` may refer to the current thread (coroutine) that the code is being called in, and threads are subject to garbage collection. After the thread (coroutine) is collected, the stored `L` value points to freed memory.

This materialized as a segfault in a complex application. It's hard to reduce to a simple test case that reliably segfaults, but running the below sample under valgrind will show the invalid memory accesses that are happening:

```lua
local uv

local f = coroutine.wrap(function ()
	uv = require "luv"
	uv.new_timer();
end);
f()
f = nil;

for i = 1, 10 do
	collectgarbage("collect")
end

uv = nil

for i = 1, 10 do
	collectgarbage("collect")
end
```

Valgrind output:

```
$ valgrind lua5.2 luv-crash.lua
==31681== Memcheck, a memory error detector
==31681== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==31681== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==31681== Command: lua5.2 luv-crash.lua
==31681== 
==31681== Invalid read of size 8
==31681==    at 0x10FAE3: lua_settop (lapi.c:165)
==31681==    by 0x64718E4: luv_close_cb (handle.c:81)
==31681==    by 0x6471D88: luv_gc_cb (handle.c:108)
==31681==    by 0x6695D7F: uv_run (in /usr/lib/x86_64-linux-gnu/libuv.so.1.0.0)
==31681==    by 0x6479361: loop_gc (luv.c:690)
==31681==    by 0x11300C: luaD_precall (ldo.c:319)
==31681==    by 0x113371: luaD_call (ldo.c:401)
==31681==    by 0x11292E: luaD_rawrunprotected (ldo.c:131)
==31681==    by 0x1135DE: luaD_pcall (ldo.c:603)
==31681==    by 0x11487E: GCTM (lgc.c:824)
==31681==    by 0x115E1D: callallpendingfinalizers (lgc.c:978)
==31681==    by 0x115E1D: luaC_freeallobjects (lgc.c:988)
==31681==    by 0x11A04D: close_state (lstate.c:226)
(...truncated...)
```

mwild1:
Some possible solutions:

[...]

PR #734, opened 2024-11-14
Use main thread of current Lua state for callbacks, when known

mwild1:
If the library is opened from a Lua thread (coroutine) that is not the main one, the stored L would be to that thread, which may become suspended (by calling yield). It is unsafe to call functions on suspended Lua threads, which luv does a lot (the stored L is used for callbacks).

The main thread of a Lua state can never yield, so it is always safe to call callbacks on.

This commit ensures the main thread is stored in the luv context instead of the current thread, when we are able to calculate it (LUA_RIDX_MAINTHREAD was added in Lua 5.2, and I have not found a way to determine the main thread in Lua 5.1).

This fixes https://github.com/luvit/luv/issues/503

### DIFF
--- a/src/luv.c
+++ b/src/luv.c
@@ -835,6 +835,15 @@ static int loop_gc(lua_State *L) {
 }
 
 LUALIB_API int luaopen_luv (lua_State* L) {
+#ifdef LUA_RIDX_MAINTHREAD
+  // Lua 5.2+ - resolve the main thread of the current Lua state, even if
+  // we were loaded from a different thread (which may become suspended/dead).
+  lua_geti(L, LUA_REGISTRYINDEX, LUA_RIDX_MAINTHREAD);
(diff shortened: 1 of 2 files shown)


Bilal2453:
So currently this does not work for PUC Lua 5.1 and for LuaJIT?

mwild1:
The behaviour will not change for Lua 5.1.

I haven't tested with LuaJIT. If it implements `LUA_RIDX_MAINTHREAD` then it should work, otherwise the behaviour will remain unchanged with this PR.

zhaozg:
Please add a test for this

mwild1:
I have added a test. It will definitely fail on Lua 5.1, and probably LuaJIT, so it is skipped there (test always passes, comment is printed).

zhaozg (approved):

Issue #735, opened 2024-11-14
Unbalanced stack after uncaught errors in work callback


mwild1:
If a work callback throws an error, luv prints it, but also prints a warning about the stack not being balanced.

Test case (reproduced with luv 1.48):

```lua
local uv = require "luv";

local function work_callback()
	error("test error");
end

local function after_work_callback()
end

local work = uv.new_work(work_callback, after_work_callback)

work:queue("hello world")

uv.run("default");
```

Output:

```shell
Uncaught Error: test error
stack traceback:
	[C]: in function 'error'
	?: in function <?:3>
	[C]: in ?
Uncaught Error: stack not balance in luv_work_cb, need 0 but 1
stack traceback:
	[C]: in ?

```

I will submit a PR with a fix shortly.

PR #736, opened 2024-11-14
luv_cfpcall: Fix stack balancing after an uncaught error

mwild1:
The default error handling code for printing uncaught errors uses luaL_tolstring(). This returns the resulting string, but also pushes it onto the stack. After we have printed it, it is safe to pop from the stack.

Leaving it on the stack causes the stack to be unbalanced.

Fixes https://github.com/luvit/luv/issues/735 (where a test case can be found)


mwild1:
I just force-pushed an updated commit, as I found a second place where luaL_tolstring() is used but the result is not removed.

zhaozg:
Please add a test to cover this PR.

mwild1:
I don't know how - do you have any suggestions? Unbalanced stack cannot be detected from Lua.

zhaozg:
> I don't know how - do you have any suggestions? Unbalanced stack cannot be detected from Lua.

That's true, ignore it.

zhaozg:
https://github.com/luvit/luv/pull/736/files#diff-d153acd02208bccdfe1c42a92c8464b423661b480f683720acff42c102ea64bbR718-L720

should do lua_pop too.

zhaozg (approved):

PR #738, opened 2024-11-17
Fix async arguments

### LINKED ISSUE #505: memory leak related to `uv.send_async`

javalikescript:
Protect async `send()` arguments using a mutex.
Use copy for async arguments rather than Lua reference.
Clear arguments in callback and before sending rather than after.
Keep existing and possibly weird behavior.

Fixes #505

This fix is rather unsatisfying and could be enhanced later.
A correct later fix would be:
* to remove async send arguments and assume that luv, as libuv, may combine multiple calls to async `send()` or
* to use a FIFO queue between async `send()` and callback ensuring the callback is executed one per send.

But both require a modification on how luv behave.

### DIFF
--- a/src/async.c
+++ b/src/async.c
@@ -16,6 +16,13 @@
  */
 #include "private.h"
 
+typedef struct {
+  luv_thread_arg_t targ;
+  uv_mutex_t mutex;
+} luv_async_arg_t;
+
+#define luv_get_async_arg_from_handle(H) ((luv_async_arg_t *) ((luv_handle_t*) (H)->data)->extra)
+
 static uv_async_t* luv_check_async(lua_State* L, int index) {
   uv_async_t* handle = (uv_async_t*)luv_checkudata(L, index, "uv_async");
   luaL_argcheck(L, handle->type == UV_ASYNC && handle->data, index, "Expected uv_async_t");
@@ -25,9 +32,19 @@ static uv_async_t* luv_check_async(lua_State* L, int index) {
 static void luv_async_cb(uv_async_t* handle) {
   luv_handle_t* data = (luv_handle_t*)handle->data;
   lua_State* L = data->ctx->L;
-  int n = luv_thread_arg_push(L, (luv_thread_arg_t*)data->extra, LUVF_THREAD_SIDE_MAIN);
-  luv_call_callback(L, data, LUV_ASYNC, n);
-  luv_thread_arg_clear(L, (luv_thread_arg_t*)data->extra, LUVF_THREAD_SIDE_MAIN);
+  luv_async_arg_t* asarg = luv_get_async_arg_from_handle(handle);
+  uv_mutex_t *argmutex = &asarg->mutex;
+  luv_thread_arg_t targcpy;
+  int n;
+  uv_mutex_lock(argmutex);
+  targcpy = asarg->targ; // work on a copy of the arguments
+  asarg->targ.argc = 0; // empty the original, nothing to clear
+  uv_mutex_unlock(argmutex);
+  n = luv_thread_arg_push(L, &targcpy, LUVF_THREAD_SIDE_MAIN);
+  if (n >= 0) {
+    luv_call_callback(L, data, LUV_ASYNC, n);
+  }
+  luv_thread_arg_clear(L, &targcpy, LUVF_THREAD_SIDE_MAIN); // clear the copy
 }
 
 static int luv_new_async(lua_State* L) {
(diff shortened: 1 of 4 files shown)


src/async.c
```diff
@@ -43,21 +113,83 @@ static int luv_new_async(lua_State* L) {
     return luv_error(L, ret);
   }
   data = luv_setup_handle(L, ctx);
-  data->extra = (luv_thread_arg_t*)malloc(sizeof(luv_thread_arg_t));
+  luv_async_arg_t* asarg = (luv_async_arg_t*)malloc(sizeof(luv_async_arg_t));
+  memset(asarg, 0, sizeof(luv_async_arg_t));
+  asarg->max = max;
+  ret = uv_mutex_init(&asarg->mutex);
+  if (ret < 0) { // unlikely
+    abort();
```
truemedian, line 121: we shouldn't be aborting the program, this should be reported as a fail

javalikescript: Fixed in commit 9fa42ef602913926b2fef0752b5d0ee8325f96e8

.github/workflows/ci.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       run: ./build/lua -e "collectgarbage('incremental', 0, 10000000000000)" tests/run.lua
 
   valgrind:
-    runs-on: ubuntu-latest
```
squeek502, line 58: Any context for this change?

.github/workflows/ci.yml
```diff
@@ -55,7 +55,7 @@ jobs:
...
+    runs-on: ubuntu-24.04
```
squeek502, line 58: Any context for this change?

javalikescript: I hardcoded the ubuntu release in an attempt to troubleshoot the CI failures, reverted in commit 090a0a76a0a3917eb3a34c443850a54eb85a34f0

javalikescript:
I added the FIFO queue because it was mentioned in the issue #505 and made it optional for the sake of compatibility.
I also wanted to keep the libuv behavior, to coalesce calls, when there is no queue nor data.

The issue is quite clear but the fix in this PR is aguable so do not hesitate if you think about any other option.

I could remove the queue from this PR, as this is a pure luv feature which will have to be supported in the long run!

Thank you for your help

PR #748, opened 2025-01-16
Bump Libuv to 1.50.0

squeek502:
https://github.com/libuv/libuv/releases/tag/v1.50.0

~~Opening as a draft because I only tested `udp_try_send2` on Windows which doesn't even support `sendmmsg` 🙃.~~


squeek502:
Windows CI is failing with

```
C:\projects\luv\deps\libuv\src\win\thread.c(298): error C4013: 'SetThreadDescription' undefined; assuming extern returning int [C:\projects\luv\build\deps\libuv\uv_a.vcxproj]
C:\projects\luv\deps\libuv\src\win\thread.c(327): error C4013: 'GetThreadDescription' undefined; assuming extern returning int [C:\projects\luv\build\deps\libuv\uv_a.vcxproj]
```

Unsure what the fix should be (or if it should be on Libuv's side or ours)

Bilal2453:
> Either one seems fine, but the second would allow us to continue to ensure that Libuv <= 1.49.2 will continue to be able to be built with MSVC2015.

that sounds a bit excessive maybe? I don't see why to continue on with Windows 8 tests/support when Libuv has already dropped it and moved on.

I want to drop Appveyor altogether frankly and merge all CI into Github Actions. Not sure if this problem would persists there, they don't use Windows Server 2012 afaik.

squeek502:
> I don't see why to continue on with Windows 8 tests/support when Libuv has already dropped it and moved on.

The point would be to ensure we match the Libuv [supported platforms](https://github.com/libuv/libuv/blob/v1.x/SUPPORTED_PLATFORMS.md), which still lists >= MSVC 2015. That is, even though Windows >= 10 is now required, MSVC >= 2015 is still documented as supported.

So, the goal would just be to test to ensure that we can always still build with MSVC 2015.

- Using the Appveyor image, but building against Libuv 1.49.2, would get us partway there, but any >= 1.50.0 specific code would not be tested
- None of the [GitHub runner images](https://github.com/actions/runner-images) seem to have MSVC 2015 installed AFAICT

So, for a full solution, we'd likely have to deal with setting up MSVC 2015 on whatever runner we go with. Ultimately, it's probably more trouble than it's worth.

Bilal2453:
> So, for a full solution, we'd likely have to deal with setting up MSVC 2015 on whatever runner we go with. Ultimately, it's probably more trouble than it's worth.

I don't think it is *that* bad? The Windows runners use Chocolatey as the package manager, and I believe there is a package for VS 2015 ready for use. Would have to test this later, for now dropping MSVC2015 is probably fine until we figure it out

zhaozg (approved):

PR #752, opened 2025-01-24
Clean up test step and exclude it from the "all" target

squeek502:
Addresses https://github.com/luvit/luv/pull/733#issuecomment-2610557444


zhaozg (approved):

clason:
Great, that indeed fixes it, thank you! Will you make a 1.50.0-1 release with this fix?

clason:
Yeah, sorry, didn't see those before commenting. That absolutely makes sense. (Tests found another leak anyway, so I couldn't update to master even if I wanted now...)

Issue #751, opened 2025-01-23
Some tests of 1.50.0 are failing on i586 and s390x


cryptomilk:
I've packaged 1.50.0.0 for Fedora and one test fails on i586 on rawhide:

```
ok 112 thread - test thread sleep msecs in main thread
Uncaught Error in thread: thread:1: syntax error near '<\21>'
  { delay = 100, elapsed = 1 }
  ./tests/test-thread.lua:87: elapsed should be at least delay 
  stack traceback:
  	[C]: in function 'assert'
  	./tests/test-thread.lua:87: in field 'fn'
  	./lib/tap.lua:59: in function <./lib/tap.lua:48>
  	[C]: in function 'xpcall'
  	./lib/tap.lua:48: in upvalue 'run'
  	./lib/tap.lua:146: in function 'lib/tap'
  	tests/run.lua:23: in main chunk
  	[C]: in ?
not ok 113 thread - test thread create with options table
```

on s390x I see:

```
  ./tests/test-misc.lua:233: assertion failed!
  stack traceback:
  	[C]: in function 'assert'
  	./tests/test-misc.lua:233: in function 'fn'
  	./lib/tap.lua:59: in function <./lib/tap.lua:48>
  	[C]: in function 'xpcall'
  	./lib/tap.lua:48: in function 'run'
  	./lib/tap.lua:146: in function 'tap'
  	tests/run.lua:23: in main chunk
  	[C]: ?
not ok 79 misc - uv.wtf8 and utf16 conversion
  ./tests/test-misc.lua:246: assertion failed!
  stack traceback:
  	[C]: in function 'assert'
  	./tests/test-misc.lua:246: in function 'fn'
  	./lib/tap.lua:59: in function <./lib/tap.lua:48>
  	[C]: in function 'xpcall'
  	./lib/tap.lua:48: in function 'run'
  	./lib/tap.lua:146: in function 'tap'
  	tests/run.lua:23: in main chunk
  	[C]: ?
not ok 80 misc - uv.wtf8<->utf16 unpaired surrogate
```

https://koji.fedoraproject.org/koji/taskinfo?taskID=128366040

Bilal2453:
The failure in `luv.sleep` is likely a regression in libuv 1.50, since that code haven't been touched on our side.

squeek502:
> Uncaught Error in thread: thread:1: syntax error near '<\21>'

This is a strange error. I'm unable to reproduce that particular error even when inserting a NAK control character into the thread function:

[...]

PR #753, opened 2025-01-24
test-misc: Allow big endian WTF-16, too

squeek502:
Addresses part of https://github.com/luvit/luv/issues/751


truemedian (approved):
Perhaps its worth mentioning in the documentation that the endian of WTF16 is the native endian of the platform as opposed to always little or big endian?

cryptomilk:
This fixes running the tests on s390x: https://koji.fedoraproject.org/koji/taskinfo?taskID=128411469

PR #772, opened 2025-05-15
docs: move to new directory

Bilal2453:
it is starting to make the root dir clog up

note: changes `require` to `dofile` because that makes more sense


squeek502:
cc @lewis6991

truemedian (changes requested):
`.gitattributes` needs to be updated with the new docs.md and meta.lua location

`README.md` needs to link to the new docs.md location

Bilal2453:
oops missed those + the make_rockspec ci, it is more scattered than I expected.

truemedian (approved):

Issue #715, opened 2024-08-31
sleep in "test threadpool with async" is non-deterministic


jamessan:
Debian's [CI for arm64](https://ci.debian.net/packages/l/lua-luv/unstable/arm64/) has recently started failing this test more often than not.

I increased the sleep to 100 and some local tests show that to be more reliable, but it would be better if there were a deterministic way to know when the async callback has been run.

I was thinking of a semaphore, but I noticed that uv API doesn't appear to be exposed in luv. Open to ideas, but for now I'll probably patch the test to use a longer sleep.

truemedian:
With https://github.com/luvit/luv/pull/774 merged using a semaphore to synchronize thread tests is now fair game.

PR #775, opened 2025-05-18
fix(test-work): use semaphore, instead of arbitrary sleep, to know when async work is done

jamessan:
On slow / busy systems, it may take longer than 10ms for the async
callback to run. When this happens, the test process typically crashes
since the async context has already been freed by the time it finally
gets scheduled to run.

Using a semaphore ensures the work is done before continuing through the
test.

Closes: #715

### DIFF
--- a/tests/test-work.lua
+++ b/tests/test-work.lua
@@ -60,34 +60,36 @@ return require('lib/tap')(function (test)
 
   test("test threadpool with async", function(print,p,expect,_uv)
     local ctx, async
-    async = _uv.new_async(expect(function (a,b,c)
+    local sem = _uv.new_sem(0)
+    async = _uv.new_async(expect(function (a,b,c,sem)
       p('in async notify callback')
       p(a,b,c)
       assert(a=='a')
       assert(b==true)
       assert(c==250)
+      sem:post()
     end))
 
     ctx = _uv.new_work(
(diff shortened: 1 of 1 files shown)


tests/test-work.lua
```diff
@@ -60,34 +60,37 @@ return require('lib/tap')(function (test)
...
     ctx = _uv.new_work(
-      function(n, s, a)         --work,in threadpool
+      function(n, s, a, sem)    --work,in threadpool
           local uv = require('luv')
           local t = tostring(uv.thread_self())
           if a then
-            assert(uv.async_send(a,'a',true,250)==0)
+            assert(uv.async_send(a,'a',true,250,sem)==0)
           end
           uv.sleep(10)
```
squeek502, line 80: Would it now be possible to remove this sleep?

jamessan: Yeah.

truemedian (changes requested):
This not so obviously deadlocks because you have a `sem_wait()` on a semaphore with a value of zero that is waiting for a `sem_post()` on the same thread.

The "right" solution here is probably a coroutine version of a semaphore, where it yields if the async request has not yet completed (and the async callback resumes the thread) or otherwise just continues

jamessan:
That's outside my experience level with Lua and this project. :) I'll leave it to you all to handle.

PR #779, opened 2025-07-28
fix: handle return value of `uv_pipe_connect2`

seandewar:
The missing check meant that no error was propagated when it returned `UV_EINVAL`. This notably made specifying `UV_PIPE_NO_TRUNCATE` ineffective.

This was likely missed out because libuv's documented return value for `uv_pipe_connect2` incorrectly stated that it returns `void`, not `int`; that's since been fixed. (libuv/libuv#4847)


squeek502:
Thanks! From looking at the libuv source, `uv_pipe_connect` just calls `uv_pipe_connect2`, so it can also error (not via `UV_PIPE_NO_TRUNCATE`, but in all the other possible ways).

Mind making the same change to `luv_pipe_connect`?

seandewar:
`uv_pipe_connect` returns `void`, so it instead sends the error to the callback. (I think that means luv would see it on the next event-loop tick?)

squeek502:
You're absolute right, my mistake.

squeek502 (approved):

PR #780, opened 2025-09-19
docs: document nil callback when needed

pynappo:
Fixes incorrect uv.fs_opendir documentation. It seem to be the only luv function where the optional callback is not at the end of the signature, so the callback param was incorrectly omitted before.


Bilal2453:
(auto-merge failed due to a failing build, unrelated failure)

squeek502:
Thanks!

CI failure is unrelated and is being tracked here: https://github.com/luvit/luv/issues/781

PR #785, opened 2025-12-11
prevent luv from trying to write zero buffers.

### LINKED ISSUE #784: Spurious segfault on empty write `pipe:write({},_)`
```bash
luajit ./minimal.lua
```

```lua
--./minimal.lua
local uv = require("luv")
local pipe = assert(uv.new_pipe())

---@diagnostic disable-next-line:missing-fields
uv.spawn("cat", {
  stdio = {
    pipe,
  },
}, function(_) end)

for i = 1, 10, 1 do
  pipe:write({}, function()
    print(i)
  end)
end

uv.run()
```

Initially discovered in https://github.com/neovim/neovim/issues/36913

From a short analysis of the code, I assume that is caused by a missing `0` check of `count` at one of
- https://github.com/luvit/luv/blob/ed1e812d82e4cf3a7363fb069c7dbddaf65309e0/src/stream.c#L162-L164
- https://github.com/luvit/luv/blob/ed1e812d82e4cf3a7363fb069c7dbddaf65309e0/src/misc.c#L94-L95
- https://github.com/luvit/luv/blob/ed1e812d82e4cf3a7363fb069c7dbddaf65309e0/src/misc.c#L60-L61

Regardless calling `malloc` on 0, seems not the best idea according to [`malloc`](https://en.cppreference.com/w/c/memory/malloc)
> If size is zero, the behavior of `malloc` is implementation-defined. For example, a null pointer may be returned. Alternatively, a non-null pointer may be returned; but such a pointer should not be [dereferenced](https://en.cppreference.com/w/c/language/operator_member_access.html), and should be passed to [free](https://en.cppreference.com/w/c/memory/free.html) to avoid memory leaks.

truemedian:
Libuv is inconsistent about handling this, in some functions it returns UV_EINVAL, in others it asserts >= 1 buffer.

To save us the effort of dealing with these inconsistencies, we simply check for zero buffers in luv and return UV_EINVAL ourselves.

Fixes #784

### DIFF
--- a/src/fs.c
+++ b/src/fs.c
@@ -579,6 +579,11 @@ static int luv_fs_write(lua_State* L) {
   req->data = luv_setup_req(L, ctx, ref);
   size_t count;
   uv_buf_t* bufs = luv_check_bufs(L, 2, &count, (luv_req_t*)req->data);
+  if (count == 0) {
+    free(bufs);
+    return luv_error(L, UV_EINVAL); // must write at least one buffer
+  }
+
   int nargs;
   FS_CALL_NORETURN(uv_fs_write, req, file, bufs, count, offset);
   free(bufs);
--- a/src/stream.c
+++ b/src/stream.c
@@ -161,6 +161,12 @@ static int luv_write(lua_State* L) {
   req->data = (luv_req_t*)luv_setup_req(L, ctx, ref);
   size_t count;
   uv_buf_t* bufs = luv_check_bufs(L, 2, &count, (luv_req_t*)req->data);
+  if (count == 0) {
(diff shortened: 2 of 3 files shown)


SinisterRectus (approved):

truemedian:
I guess it mostly depends on how we want to classify this error; if it's always the programmer's fault (in the same class as passing the wrong type) then throwing an error would probably make sense.

However, in this case I was treating it as just an invalid parameter. Mostly just boils down to whether we think this case should throw or just return the fail triplet. Would appreciate feedback on which feels more correct.

PR #788, opened 2025-12-30
Add support for and test against Lua 5.5

### LINKED ISSUE #787: Support for Lua 5.5
https://www.lua.org/manual/5.5/readme.html#changes

squeek502:
Intended to close #787

However, Lua 5.5.0 seems to somehow introduce regressions in our `thread` and `work` tests. Some of the thread tests hang, and the `test threadpool` test of `test-work` fails with

```
Uncaught Error: ./tests/test-work.lua:24: attempt to perform arithmetic on a nil value (local 'n')
stack traceback:
        ./tests/test-work.lua:24: in function <./tests/test-work.lua:23>
        [C]: in field 'run'
        ./lib/tap.lua:61: in function <./lib/tap.lua:48>
        [C]: in global 'xpcall'
        ./lib/tap.lua:48: in upvalue 'run'
        ./lib/tap.lua:146: in local 'tap'
        tests/run.lua:23: in main chunk
        [C]: in ?
```

Need to figure out what's going on. The [documented incompatibilities](https://www.lua.org/manual/5.5/manual.html#8) don't seem relevant on first glance.

### DIFF
--- a/.ci/winmake.bat
+++ b/.ci/winmake.bat
@@ -154,33 +154,56 @@ IF NOT EXIST %LUA_H% (
    goto :EXITERROR
 )
 
-findstr /R /C:"#define[ %TABCHAR%][ %TABCHAR%]*LUA_VERSION_MAJOR"  %LUA_H% > NUL
-if NOT %ERRORLEVEL%==0 (
-   rem ECHO We've got a Lua version 5.1
-   rem findstr /R /C:"#define[ %TABCHAR%][ %TABCHAR%]*LUA_VERSION[ %TABCHAR%]"  %LUA_H%
-   SET LUA_VER=5.1
-) else (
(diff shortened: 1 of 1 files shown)


truemedian:
The list of incompatibilities isn't very descriptive, but the clang-asan test fails that we're getting a null string from dumping the function and the rest of the tests are failing as if the thread didn't start with the function we expect.

So this seems to have been a change with respect to how lua_dump works.

The implementation of `string.dump` in Lua 5.5 has an additional branch that we likely need to reproduce: <https://www.lua.org/source/5.5/lstrlib.c.html#writer>

squeek502:
Thanks, your comment got me pointed in the right direction. `luv_thread_dumped` should now be much more resilient against future changes to the details of `lua_dump` as well.

PR #794, opened 2026-02-03
Allow using WITH_SHARED_LIBUV with rockspec build

tobil4sk:
This makes it possible to use the system libuv when installing using the rockspec


squeek502:
Thanks!

PR #796, opened 2026-02-12
Bump to Libuv 1.52.0

### LINKED ISSUE #792: vim.uv.os_uname().info.machine returns 'unknown' for Windows ARM CPU
# Problem
When running Neovim with Windows OS on a ARM Processor, the machine info is unknown.

vim.uv.os_uname().info.machine
Steps to reproduce
Install Neovim (nightly or stable) on running Windows with an ARM Processor.

I have tried to install `luv` via `luarocks` but it is failing to build. I am not actually able to run `luv` and run `uv.os_uname()`. 

-- init.lua
vim.print(vim.uv.os_uname())
Expected behavior
machine should have the correct value

Nvim version (nvim -v)
v0.12.0-dev-1998+g272ec9627c

Operating system/version
Windows 11

Terminal name/version
Windows Terminal 1.23.13403.0

$TERM environment variable
n/a

Installation
scoop install neovim-nightly

I opened an [issue ](https://github.com/neovim/neovim/issues/37329#issuecomment-3746776950) on Neovim, but this looks to be an issue with `luv` and not `neovim`.

### LINKED ISSUE #795: Update for Libuv v1.52.0
https://github.com/libuv/libuv/releases/tag/v1.52.0

truemedian:
https://github.com/libuv/libuv/releases/tag/v1.52.0

+ Adds `udp_open_ex`, which allows passing `reuseaddr` and `reuseport` options to `uv_udp_open`.
+ Adds `tcp_keepalive_ex`, which allows passing `KEEPINTVL` and `KEEPCNT` options to `uv_tcp_keepalive`.
+ Adds the `frsize` field to the statfs table, which contains the fragment size of the filesystem.
+ Adds the `UDP_REUSEPORT` constant to the constant table.

The existing suite has no tests for `udp_open`, so I have not added any for `udp_open_ex`. Given that `uv_udp_open` just calls `uv_udp_open_ex` with default options, this change shouldn't cause any regressions.


zhaozg (approved):

truemedian:
My thought process was to ensure that the user could expect the semantics of any flags that they pass to luv (ie. if you pass reuseport, then you can expect that the handle has those semantics; or libuv throws EINVAL because the system doesn't support it), but I see now that the other cases of `_ex`  functions instead just drop whatever flags the function doesn't support.

Will go ahead and draft up a fix to merge both functions back into their non `_ex` form.

PR #797, opened 2026-02-13
Ensure consistent handling of _ex functions.

truemedian:
The existing behavior around `_ex` functions has been to coalesce them with their other forms and silently ignore flags that are not supported on the target libuv version. I got this wrong in the last PR, so here's to making that consistent.

+ `udp_open_ex` has been merged into `udp_open` and handles the given flags when possible.
+ `tcp_keepalive_ex` has been merged into `tcp_keepalive` and handles the given flags when possible.


squeek502 (approved):
Looks good, thanks!

PR #799, opened 2026-02-24
assorted bug fixes and many memory corruption fixes

truemedian:
This is a lot of changes all rolled into one because it would otherwise be 30 different single-line commits.

Most of the issues are related to not checking the return of malloc for a null return.


src/process.c
```diff
@@ -64,7 +64,7 @@ static int sparse_rawlen(lua_State* L, int tbl) {
   tbl = lua_absindex(L, tbl);
 
   lua_pushnil(L);
-  while (lua_next(L, -2)) {
+  while (lua_next(L, tbl)) {
```
squeek502: This whole `sparse_rawlen` function seems like it could just be deleted. With this fix, we're at best saving a few bytes by not just allocating `[3]uv_stdio_container_t` every time, and at worst over-allocating when the `stdio` table has more than 3 integer keys.

It also seems like we might not even need heap allocation for `options.stdio`?

Really, this whole stdio related code seems a bit convoluted/buggy, so maybe a separate PR focusing on cleaning it up would be warranted.

src/util.h
```diff
@@ -25,6 +25,4 @@
 #define LUV_UV_VERSION_LEQ(major, minor, patch) \
   (((major)<<16 | (minor)<<8 | (patch)) >= UV_VERSION_HEX)
 
-void luv_stack_dump(lua_State* L, const char* name);
```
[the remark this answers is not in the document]
truemedian: Removed mostly just because it's never used, if it's a helpful debugging tool I can put it back

src/process.c
truemedian: I wasn't trying to outright remove anything, but I do agree that stdio (really all of spawn's handling) could be improved.

squeek502 (approved):
Nice work!

PR #806, opened 2026-02-28
udp: Add `bind` flags that got lost along the way

squeek502:
These flags were added in 1.42.0 and 1.49.0 but went unnoticed.


zhaozg (approved):
VIP

PR #810, opened 2026-03-03
Use a unique string for the handle registry instead of a reference

### LINKED ISSUE #808: regression in uv.walk
Take the following reproduction code:

```lua
local uv = require 'uv'
local t = uv.new_tty(0, true)

coroutine.wrap(function()
        uv.walk(print)
end)()

uv.run()
```
Before commit https://github.com/luvit/luv/commit/3ff02c8c2beaaf125f3ee5db950138913ffcf978 it used to work fine (printing the tty handle), after this commit, the `uv.walk` call now raises an unexpected error:
```
$ luajit main.lua
Uncaught Error: attempt to call a table value
stack traceback:
	[builtin#37]: at 0x5648b5928b80
	main.lua:8: in main chunk
	[C]: at 0x5648b58d46b0
```

Not entirely sure what's happening here. but I was able to blame it on this commit after the issue was reported in https://github.com/luvit/lit/issues/357.

Note: the coroutine is important, doesn't happen without it.

truemedian:
There can only be one luv_context per lua registry, so there can also only be one handle registry per lua registry. Therefore, we don't need to store this as a reference; we can use a unique registry key.

This allows simplifying luv_walk to pass only the lua_State again, which is necessary to fix #808. The existing code uses the loop's Lua state, but luv_walk_cb is always called synchronously and expects the arguments of the function to be accessible on the stack.

### DIFF
--- a/src/loop.c
+++ b/src/loop.c
@@ -95,31 +95,35 @@ static int luv_update_time(lua_State* L) {
 }
 
 static void luv_walk_cb(uv_handle_t* handle, void* arg) {
-  luv_ctx_t* ctx = (luv_ctx_t*)arg;
-  lua_State* L = ctx->L;
+  lua_State* L = (lua_State*)arg;
   luv_handle_t* data = (luv_handle_t*)handle->data;
 
   // Skip foreign handles (shared event loop)
-  lua_rawgeti(L, LUA_REGISTRYINDEX, ctx->ht_ref);
-  lua_rawgetp(L, -1, data);
+  lua_rawgetp(L, 2, data);
   if (lua_isnil(L, -1)) {
-    lua_pop(L, 2);
+    lua_pop(L, 1);
     return;
   }
+  lua_pop(L, 1);
 
-  // Sanity check
-  // Most invalid values are large and refs are small, 0x1000000 is arbitrary.
-  assert(data && data->ref < 0x1000000);
+  // We know that data is a valid luv_handle_t* because the above check succeeded.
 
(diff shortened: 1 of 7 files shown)


src/loop.c
```diff
@@ -95,31 +95,35 @@ static int luv_update_time(lua_State* L) {
...
 static void luv_walk_cb(uv_handle_t* handle, void* arg) {
-  luv_ctx_t* ctx = (luv_ctx_t*)arg;
-  lua_State* L = ctx->L;
+  lua_State* L = (lua_State*)arg;
   luv_handle_t* data = (luv_handle_t*)handle->data;
 
   // Skip foreign handles (shared event loop)
-  lua_rawgeti(L, LUA_REGISTRYINDEX, ctx->ht_ref);
-  lua_rawgetp(L, -1, data);
+  lua_rawgetp(L, 2, data);
```
squeek502: Might be good to have a comment explaining the `2` here, or maybe a comment at the start of the function explaining the assumed state of the Lua stack (if I understand correctly, index 1 is expected to be the user-provided callback function, and index 2 is expected to be `registry["LUV_HANDLES"]`?).

Maybe also a comment somewhere mentioning something about "luv_walk_cb is always called synchronously and expects the arguments of the function to be accessible on the stack" from the PR description.

squeek502: Perfect, thanks!

squeek502 (approved):

zhaozg (approved):

PR #813, opened 2026-03-11
Use assert on bind (on the TCP echo server example)

nilFinx:
This will make sure that people will wrap listen part with something, to ensure that nobody is getting confused when the server appears to start up fine, but it actually didn't.


squeek502:
`docs/docs.md` is generated, please edit `docs/docs.lua` instead and run `lua docs/docgen.lua`

nilFinx:
I apologize for the mistake, it has been fixed now.

truemedian (approved):
CI failures are unrelated. Seems to be GitHub Actions acting up.

PR #817, opened 2026-03-18
Add POS36-C runtime safety check to setuid and security documentation

### LINKED ISSUE #341: Library calls setuid and/or setgid without setgroups or initgroups
When building package for openSUSE, rpmlint warns me:

> This executable is calling setuid and setgid without setgroups or initgroups. There is a high probability this means it didn't relinquish all groups, and this would be a potential security issue to be fixed. Seek POS36-C on the web for details about the problem.

mcepl:
Add a runtime check in `luv_setuid()` that prevents dropping from root (`euid 0`) to an unprivileged user while supplementary groups are still set. This addresses the CERT POS36-C vulnerability where `setuid()` alone does not revoke supplementary group privileges.

The check returns an error instructing users to call `uv.setgroups({})` or `uv.initgroups()` before `uv.setuid()`. Non-root callers and
root-to-root transitions are unaffected.

Also adds security warnings to `setuid`/`setgid` documentation in all
three doc files (`docs.md`, `docs.lua`, `meta.lua`) explaining the correct
privilege-dropping order.

See: https://wiki.sei.cmu.edu/confluence/display/c/POS36-C
Fixes: https://github.com/luvit/luv/issues/341

### DIFF
--- a/src/misc.c
+++ b/src/misc.c
@@ -477,7 +477,33 @@ static int luv_getgid(lua_State* L){
 
 static int luv_setuid(lua_State* L){
   int uid = luaL_checkinteger(L, 1);
-  int r = setuid(uid);
+  int r;
+
+  /* POS36-C safety check: when dropping privileges from root to an
+     unprivileged user, supplementary groups must be dropped first.
+     See: https://wiki.sei.cmu.edu/confluence/display/c/POS36-C */
+  if (geteuid() == 0 && uid != 0) {
+    if (getgid() == 0 || getegid() == 0) {
+      return luaL_error(L,
+        "Cannot setuid before dropping group privileges "
+        "(POS36-C security check). Call uv.setgid() before "
+        "dropping privileges with uv.setuid().");
+    }
+
+    int ngroups = getgroups(0, NULL);
+    if (ngroups == -1) {
+      return luaL_error(L, "Error getting number of groups");
+    }
+
+    if (ngroups > 0) {
+      return luaL_error(L,
+        "Cannot setuid while supplementary groups are still set "
+        "(POS36-C security check). Call uv.setgroups({}) or "
+        "uv.initgroups() before dropping privileges with uv.setuid().");
+    }
+  }
+
+  r = setuid(uid);
   if (-1 == r) {
     luaL_error(L, "Error setting UID");
   }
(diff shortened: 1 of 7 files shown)


truemedian:
`docs.md` and `meta.lua` are both generated files from the definitions in docs.lua. You need to run `docs/docgen.lua` to update the files for CI to pass.

mcepl:
> `docs.md` and `meta.lua` are both generated files

So, they shouldn't be in git at all?

Bilal2453:
Where else would they be 😆

src/misc.c
```diff
@@ -491,6 +519,82 @@ static int luv_setgid(lua_State* L){
...
+    return luaL_error(L, "Error getting number of groups");
+  }
+
+  groups = NULL;
+  if (ngroups > 0) {
+    groups = (gid_t*)malloc(ngroups * sizeof(gid_t));
+  }
+  if (ngroups > 0 && !groups) {
+    return luaL_error(L, "Error allocating memory for groups");
+  }
```
Bilal2453, line 538:
```suggestion
  groups = NULL;
  if (ngroups > 0) {
    groups = (gid_t*)malloc(ngroups * sizeof(gid_t));
    if (!groups) {
      return luaL_error(L, "Error allocating memory for groups");
    }
  }
```

Why repeat the `ngroups` check?

src/misc.c
```diff
@@ -491,6 +519,82 @@ static int luv_setgid(lua_State* L){
...
+
+static int luv_setgroups(lua_State* L) {
+  int ngroups, i;
+  gid_t* groups;
+
+  luaL_checktype(L, 1, LUA_TTABLE);
+  ngroups = lua_rawlen(L, 1);
+
+  groups = (gid_t*)malloc(ngroups * sizeof(gid_t));
+  if (!groups && ngroups > 0) {
```
Bilal2453, line 568: If there are less than 1 entries, you can just return early here.

mcepl: Like this https://github.com/luvit/luv/pull/817/commits/721ba963e2a95b862dfd368d3f6da4006855bc9c ?

src/misc.c
mcepl: Done

PR #820, opened 2026-03-30
ci: Update aggressive GC settings for Lua 5.5, fix potential use-after-frees

squeek502:
The previous call only switches to the incremental GC in Lua 5.5; the numbers afterwards are the Lua 5.4-specific way of setting the pause and stepmul.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,9 @@ jobs:
       run: make test
     - name: Test with Aggressive GC
       # Run with extremely aggressive garbage collection to potentially find more problems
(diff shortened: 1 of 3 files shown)


squeek502:
Wasn't expecting this to find actual failures 🙃 

Looks to be scandir related

EDIT: This is actually finding quite a bit. I guess previous Lua versions never really let us tune the GC to be *this* aggressive (Lua 5.5 allows "producing a non-incremental, stop-the-world collector"). Or we just didn't tune it as aggressively as possible.

truemedian (approved):

PR #822, opened 2026-03-30
breaking change: Remove ad-hoc setuid/setgid bindings

### LINKED ISSUE #341: Library calls setuid and/or setgid without setgroups or initgroups

squeek502:
These functions aren't coming from libuv, they aren't portable (to the point of the function bindings being wrapped in a `#ifndef _WIN32`), and there's proper support for setgid/setuid in the libuv process spawning API.

These functions are also [seemingly unused except by a single personal project for which the bindings were added in the first place](https://github.com/luvit/luv/issues/341#issuecomment-4093855494). The ad-hoc getuid/getgid functions do see some use in neovim plugins, so they remain but it's now recommended to use the proper `os_get_passwd` function from libuv instead.

There's also some potential security issues around setuid/setgid that we aren't dealing with correctly and there's no good reason to take on that responsibility (closes https://github.com/luvit/luv/issues/341).

---

Alternative to https://github.com/luvit/luv/pull/817

### DIFF
--- a/docs/docs.lua
+++ b/docs/docs.lua
@@ -4442,7 +4442,12 @@ local doc = {
           desc = 'Returns the user ID of the process.',
           returns = 'integer',
           notes = {
-            'This is not a libuv function and is not supported on Windows.',
+            'This is not a libuv function and does not exist when targeting Windows.',
+          },
+          warnings = {
+            [[
+              Deprecated. Use `uv.os_get_passwd()` instead.
+            ]],
           },
         },
         {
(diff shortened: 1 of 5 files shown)


docs/docs.lua
```diff
@@ -4442,35 +4442,25 @@ local doc = {
           desc = 'Returns the user ID of the process.',
           returns = 'integer',
           notes = {
-            'This is not a libuv function and is not supported on Windows.',
+            'This is not a libuv function and does not exist when targeting Windows.',
+          },
+          warnings = {
+            [[
+              This is a legacy function that may be removed in the future. It's recommended to use `uv.os_get_passwd()` instead.
```
clason:
```suggestion
              Deprecated. Use `uv.os_get_passwd()` instead.
```

I would phrase this even more strongly.

Bilal2453: It is, as far as I can tell, and by all means, a security threat at best, or an active vulnerability at worst, as I can understand? so yeah I agree

clason: You can also add _when_ (i.e., what luv version) it will be removed.

Bilal2453: Frankly, if not for the Neovim usage, I believe we would have removed it as soon as we could.

clason: We break enough stuff ourselves, don't worry too much on our account ;) The alternative is also easy to switch to and for us to document. 

(But I appreciate the thought!)

I think it's good to have a full version where users can switch, i.e., introduce the warning for the 1.53 release and remove at 1.54. If you want to fast-track it, make this 1.52.2 and 1.53.0, respectively.

squeek502: @Bilal2453 to be clear, there's no security issue with the *get* functions AFAIK, that only pertains to the *set* functions which are removed entirely in this PR.

I think "deprecated" is probably a better description for the *get* functions.

truemedian (approved):
I'm more on board with removing these than with trying to ensure people don't use them incorrectly.

PR #824, opened 2026-04-18
fix: make luv_sig_string_to_num case insensitive

AlexCodesApps:
This commit allows for passing signals in all caps like "SIGINT" and
mix-cased. The behavior is arguably better and more intuitive as the
current behavior is to silently fails if the input string is not
in all lower case. Also adds utility functions for comparing case
insensitive strings portable without locale.

### DIFF
--- a/src/constants.c
+++ b/src/constants.c
@@ -519,8 +519,7 @@ static int luv_strcaseequal(const char *s1, const char *s2) {
     s1++;
     s2++;
   }
-  if (*s1 != *s2)
-    return 0;
+  return *s1 == *s2;
 }
 
 static int luv_sig_string_to_num(const char* string) {
--- a/src/constants.c
+++ b/src/constants.c
@@ -501,112 +501,134 @@ static const char* luv_sock_num_to_string(const int num) {
   return NULL;
 }
 
+static int luv_isalpha(int c) {
+  return ('a' <= c && c <= 'z')
+      || ('A' <= c && c <= 'Z');
+}
+
+static int luv_tolower(int c) {
+  if (luv_isalpha(c))
+    return c | 0x20;
+  return c;
+}
+
+static int luv_strcaseequal(const char *s1, const char *s2) {
+  while (*s1 && *s2) {
+    if (luv_tolower(*s1) != luv_tolower(*s2))
+      return 0;
+    s1++;
+    s2++;
+  }
+  if (*s1 != *s2)
+    return 0;
+}
+
 static int luv_sig_string_to_num(const char* string) {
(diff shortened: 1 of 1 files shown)


squeek502:
> the current behavior is to silently fails if the input string is not in all lower case

Can you give an example of this? The failure does not seem silent:

https://github.com/luvit/luv/blob/a9f2e5d5bb2f97fb1be2543266b89dd6c61d152f/src/signal.c#L51-L52

src/constants.c
```diff
@@ -501,112 +501,134 @@ static const char* luv_sock_num_to_string(const int num) {
...
+static int luv_strcaseequal(const char *s1, const char *s2) {
+  while (*s1 && *s2) {
+    if (luv_tolower(*s1) != luv_tolower(*s2))
+      return 0;
+    s1++;
+    s2++;
+  }
+  if (*s1 != *s2)
+    return 0;
+}
```
squeek502, line 523: This function only returns 0

AlexCodesApps:
> > the current behavior is to silently fails if the input string is not in all lower case
> 
> Can you give an example of this? The failure does not seem silent:
> 
> https://github.com/luvit/luv/blob/a9f2e5d5bb2f97fb1be2543266b89dd6c61d152f/src/signal.c#L51-L52

it appears that snippet of code is missing in `luv_process_kill`. Perhaps I can fix just that?

src/constants.c
AlexCodesApps: corrected this in a new commit

squeek502:
Sure, could make that a separate PR if you want since the case insensitivity change might be worthwhile, too.

I'd add the check here:

https://github.com/luvit/luv/blob/a9f2e5d5bb2f97fb1be2543266b89dd6c61d152f/src/process.c#L312

Something like:

```c
int signum = luv_sig_string_to_num(lua_tostring(L, slot));
luaL_argcheck(L, signum, slot, "Invalid Signal name");
return signum;
```

AlexCodesApps:
> Sure, could make that a separate PR if you want since the case insensitivity change might be worthwhile, too.
> 
> I'd add the check here:
> 
> https://github.com/luvit/luv/blob/a9f2e5d5bb2f97fb1be2543266b89dd6c61d152f/src/process.c#L312
> 
> Something like:
> 
> ```c
> int signum = luv_sig_string_to_num(lua_tostring(L, slot));
> luaL_argcheck(L, signum, slot, "Invalid Signal name");
> return signum;
> ```

done. https://github.com/luvit/luv/pull/825

PR #825, opened 2026-04-18
fix: add validation to luv_parse_signal

AlexCodesApps:
Adds validation of signals passed as strings to luv_parse_signal, and
transitively to luv_process_kill and luv_kill. The previous behavior was
to swallow errors from invalid signals.


squeek502 (approved):
Thanks!

